PR #11, opened 2026-04-16
refactor: Implement BaseHoverPopup for ContextWindowPopup

ethanyhou:
In the coming PR to build the hover pop for quota usage pie chart, the popup style is same as the context window popup：
image
So I extracted a base hover popup class for the future quota popup to reuse.

**Test:**
The context window popup should work as before:
image


jdneo (approved):
LGTM

PR #5, opened 2026-04-13
feat: [New billing] Add Usage Status panel with to the preferences.

ethanyhou:
**Test:** 

1. **Accounts:**
- Pro: xcode-pro
- ProPlus: xcode-proplus
- Max: eclipse-proplus
- CB: eclipse-cb
- CE: MS account but may need extra FFs
- Free: xcode-test

2. **CLS:**
- Switch to `qingyliu/models` 

3. **UI:**
**Free**
   - image

**Pro:**
   - image

**ProPlus:**
   - image

**Max:**
   - image

**CB CE without limit:**
   - image

**Not signed in:**
   - Implementation:
   - image

4. For accounts without TBB feature flag:
   - image

**Note:**
1. The rest UI are supported in this PR but cannot be tested due to no quota configuration from CAPI side. You may hardcode the CLS side and change the quota/unlimited props to check the effects in Eclipse.
2. Threshold configuration was removed. The limit bar should become yellow >= 75&, red >= 95%
3. The first button should be primary button (either update plan or manage budget button)
4. Overage label has two states: overage configured / not configured. You can set up when it can be configured in GitHub settings -> Copilot -> Features:
image

6. Rest UX in Figma: https://www.figma.com/design/WLPCfP87Y2dXDXwAvqCl4I/Token-Based-Billing?node-id=5-19&t=04KEOyc7gBekzVh1-1

PR #16, opened 2026-04-20
feat: [New billing] Implement quota usage indicator with dynamic pie chart and popup details

ethanyhou:
This PR should be merged after https://github.com/microsoft/copilot-for-eclipse/pull/5 due to dependency.

Test:
image
image


com.microsoft.copilot.eclipse.core/src/com/microsoft/copilot/eclipse/core/AuthStatusManager.java
```diff
@@ -177,6 +186,15 @@ public void setQuotaStatus(CheckQuotaResult checkQuotaResult) {
     this.checkQuotaResult = checkQuotaResult;
   }
 
+  /**
+   * Patch the cached quota result with fields from a push quota change notification.
+   *
+   * @param notification the push notification received from the language server
+   */
+  public void patchQuotaFromNotification(QuotaChangeNotification notification) {
+    getQuotaStatus().updateFromNotification(notification);
```
jdneo, line 195: Should we construct a new CheckQuotaResult instead of mutating the cached one? Mutating in place bypasses setQuotaStatus and risks UI readers (e.g. QuotaUsageIndicator) seeing a partially-updated snapshot. Something like:

```java
setQuotaStatus(CheckQuotaResult.merge(getQuotaStatus(), notification));
```

would keep setQuotaStatus as the single update path and make the swap atomic from the reader's side. WDYT?

ethanyhou:
Close this PR as pending TBB from CAPI side.

PR #26, opened 2026-04-27
In Agent mode: Auto scroll to prompts like 'Continue'.

raghucssit:
User must be made aware some action is needed from them to continue the work by agent. Usually this happens when "Too many requests" or "Command line run prompt" etc.

see https://github.com/microsoft/copilot-for-eclipse/issues/120

### DIFF
--- a/com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/BaseTurnWidget.java
+++ b/com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/BaseTurnWidget.java
@@ -23,6 +23,7 @@ import org.eclipse.swt.widgets.Label;
 import org.eclipse.ui.PlatformUI;
 import org.osgi.service.event.EventHandler;
 
+import com.microsoft.copilot.eclipse.core.Constants;
 import com.microsoft.copilot.eclipse.core.CopilotCore;
 import com.microsoft.copilot.eclipse.core.chat.ConfirmationContent;
 import com.microsoft.copilot.eclipse.core.events.CopilotEventConstants;
@@ -36,6 +37,7 @@ import com.microsoft.copilot.eclipse.core.persistence.CopilotTurnData;
 import com.microsoft.copilot.eclipse.core.persistence.CopilotTurnData.EditAgentRoundData;
 import com.microsoft.copilot.eclipse.core.persistence.CopilotTurnData.ReplyData;
 import com.microsoft.copilot.eclipse.core.persistence.CopilotTurnData.ToolCallData;
+import com.microsoft.copilot.eclipse.ui.CopilotUi;
 import com.microsoft.copilot.eclipse.ui.chat.services.AvatarService;
 import com.microsoft.copilot.eclipse.ui.chat.services.ChatServiceManager;
 import com.microsoft.copilot.eclipse.ui.utils.SwtUtils;
@@ -604,16 +606,16 @@ public abstract class BaseTurnWidget extends Composite {
    * @param code the server error code
    * @param modelProviderName the BYOK model-provider name, or {@code null} for built-in models
    */
-  protected void createWarnDialog(String message, int code, String modelProviderName) {
+  protected Composite createWarnDialog(String message, int code, String modelProviderName) {
     // TODO: Remove this legacy fallback after TBB is officially released.
     // When the language server has not enabled token-based billing yet, restore the original
     // main-branch warning behavior (no plan-driven actions; single upgrade button on the legacy
     // 30-day free trial message).
     if (!this.serviceManager.getAuthStatusManager().getQuotaStatus().tokenBasedBillingEnabled()) {
-      new WarnWidget(this, SWT.BOTTOM, message, code);
+      WarnWidget warnWidget = new WarnWidget(this, SWT.BOTTOM, message, code);
       ensureFooterAtBottom();
       requestLayout();
-      return;
+      return warnWidget;
     }
     boolean byokQuotaExceeded = QuotaActions.isByokQuotaExceeded(code, modelProviderName);
     String displayMessage = byokQuotaExceeded ? Messages.chat_warnWidget_byokQuotaUsageMessage : message;
@@ -627,9 +629,10 @@ public abstract class BaseTurnWidget extends Composite {
           && quotaStatus.premiumInteractions().overagePermitted();
       canUpgradePlan = quotaStatus.canUpgradePlan();
     }
-    new WarnWidget(this, SWT.NONE, displayMessage, planForActions, overageEnabled, canUpgradePlan);
+    WarnWidget warnWidget = new WarnWidget(this, SWT.NONE, displayMessage, planForActions, overageEnabled, canUpgradePlan);
     ensureFooterAtBottom();
     requestLayout();
+  return warnWidget;
   }
 
   /**
@@ -665,6 +668,19 @@ public abstract class BaseTurnWidget extends Composite {
         .getConfirmationFuture();
 
     this.getParent().requestLayout();
+
+    // Ensure the chat content viewer scrolls to show the newly created confirmation
+    // dialog. Walk up the composite hierarchy to find a ChatContentViewer
+    // and request scrolling. Use async exec because layout needs to complete first.
+    SwtUtils.invokeOnDisplayThreadAsync(() -> {
+      ChatContentViewer viewer = SwtUtils.findParentOfType(this.getParent(), ChatContentViewer.class);
+      if (viewer != null) {
+        if (this.confirmDialog != null && !this.confirmDialog.isDisposed()) {
+          viewer.showControl(this.confirmDialog);
+        }
+      }
+
+    }, this.getParent());
 
     return toolConfirmationFuture;
   }
--- a/com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/BaseTurnWidget.java
+++ b/com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/BaseTurnWidget.java
@@ -629,10 +629,11 @@ public abstract class BaseTurnWidget extends Composite {
           && quotaStatus.premiumInteractions().overagePermitted();
       canUpgradePlan = quotaStatus.canUpgradePlan();
     }
-    WarnWidget warnWidget = new WarnWidget(this, SWT.NONE, displayMessage, planForActions, overageEnabled, canUpgradePlan);
+    WarnWidget warnWidget =
+        new WarnWidget(this, SWT.NONE, displayMessage, planForActions, overageEnabled, canUpgradePlan);
     ensureFooterAtBottom();
     requestLayout();
-  return warnWidget;
+    return warnWidget;
   }
 
   /**
--- a/com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatContentViewer.java
+++ b/com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatContentViewer.java
@@ -468,9 +468,17 @@ public class ChatContentViewer extends Composite {
   }
 
   private void renderWarnMessageWithUpgradePlanButton(String errorMessage, int code, String modelProviderName) {
-    latestTurnWidget.createWarnDialog(errorMessage, code, modelProviderName);
+    Composite warnWidget = latestTurnWidget.createWarnDialog(errorMessage, code, modelProviderName);
     refreshLayoutFull();
     scrollToLatestUserTurn();
+    // Ensure the chat content viewer scrolls to show the newly created warning banner. Walk up the composite hierarchy
+    // to find a ChatContentViewer and request scrolling. Use async exec because layout needs to complete first.
+    SwtUtils.invokeOnDisplayThreadAsync(() -> {
+      if (warnWidget != null && !warnWidget.isDisposed()) {
+        showControl(warnWidget);
+      }
+
+    }, this.getParent());
   }
 
   /**
@@ -483,6 +491,12 @@ public class ChatContentViewer extends Composite {
     this.errorWidget = new ErrorWidget(cmpContent, SWT.BOTTOM, errorMessage);
     refreshLayoutFull();
     scrollToLatestUserTurn();
+    // Ensure the chat content viewer scrolls to show the newly created error banner.
+    SwtUtils.invokeOnDisplayThreadAsync(() -> {
+      if (this.errorWidget != null && !this.errorWidget.isDisposed()) {
+        this.showControl(this.errorWidget);
+      }
+    }, this.getParent());
   }
 
   /**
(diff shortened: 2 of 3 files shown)


com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatContentViewer.java
```diff
@@ -204,9 +204,17 @@ public void processTurnEvent(ChatProgressValue value) {
...
-      // Auto-scroll to bottom if enabled
-      if (shouldAutoScrollToBottom()) {
-        scrollToBottom();
+      // For agent-mode responses (agent rounds/tool calls) we always force the view
+      // to scroll to the bottom so prompts that require user action (e.g. Continue,
+      // permission dialogs) are visible even if the user previously scrolled away.
+      if (value.getAgentRounds() != null && !value.getAgentRounds().isEmpty()) {
+        // Use a forced scroll to ensure visibility regardless of manual scroll state.
+        forceScrollToBottom();
+      } else {
```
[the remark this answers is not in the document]
iloveeclipse: Check Throttler API, see https://github.com/eclipse-platform/eclipse.platform.ui/blob/22d9cd153d92d61a44f1a1f765d39ced294a4bee/bundles/org.eclipse.jface/src/org/eclipse/jface/util/Throttler.java

raghucssit:
@jdneo I have fixed the copilot review comments..
Please check this PR.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatContentViewer.java
```diff
@@ -375,12 +393,49 @@ private boolean shouldAutoScrollToBottom() {
...
+      if (this.isDisposed()) {
+        return;
+      }
+
+      Rectangle clientArea = this.getClientArea();
+      // compute content height with current client width
+      Point containerSize = cmpContent.computeSize(clientArea.width, SWT.DEFAULT);
+      int contentHeight = containerSize.y;
+      int originY = Math.max(0, contentHeight - clientArea.height);
+      this.setOrigin(0, originY);
```
jdneo: duplicated with doScrollToBottom()?

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatContentViewer.java
```diff
@@ -204,9 +214,17 @@ public void processTurnEvent(ChatProgressValue value) {
       }
       refreshScrollerLayout();
 
-      // Auto-scroll to bottom if enabled
-      if (shouldAutoScrollToBottom()) {
-        scrollToBottom();
+      // For agent-mode responses (agent rounds/tool calls) we always force the view
+      // to scroll to the bottom so prompts that require user action (e.g. Continue,
+      // permission dialogs) are visible even if the user previously scrolled away.
```
jdneo: According to this comment, it seems that changes in com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/BaseTurnWidget.java is not required?

jdneo:
@raghucssit Thank you. I left some comments.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/utils/SwtUtils.java
```diff
@@ -38,6 +38,27 @@ private SwtUtils() {
...
+   * Walks up the parent chain of the given control and returns the first ancestor that is an instance of the specified
+   * type, or {@code null} if none is found.
+   *
+   * @param <T> the target type
+   * @param control the starting control (may be {@code null})
+   * @param type the class to search for
+   * @return the first matching ancestor, or {@code null}
+   */
+  @Nullable
+  public static <T> T findParentOfType(Control control, Class<T> type) {
```
jdneo, line 64: move this below `private static final int DEFAULT_GHOST_TEXT_SCALE = 128;`

raghucssit: Fixed.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatContentViewer.java
raghucssit: We need both the fix.
This fix works for too max tool attempts confirmation and BaseTurnWidget fix works to scroll when tool run confirmation is asked.
`max tool attempts `
Image
`tool confirmation`
Image

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatContentViewer.java
raghucssit: Yes.. Cleaned up..

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatContentViewer.java
```diff
@@ -204,9 +214,17 @@ public void processTurnEvent(ChatProgressValue value) {
...
+      if (value.getAgentRounds() != null && !value.getAgentRounds().isEmpty()) {
```
jdneo, line 269: Is there any reason to force scroll to bottom always when new agent round comes? Even if the user scroll is up (when scroll bar is not at the bottom)?

iloveeclipse:
@raghucssit: there are checkstyle errors, please check https://github.com/microsoft/copilot-for-eclipse/actions/runs/25695468318/job/75537241263?pr=26

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatContentViewer.java
raghucssit: Yes. I agree. I tested once again. force scroll happens for all agent responses.
I am trying to fix it.
If user has scrolled up for reading the response then only in case of confirmation dialog/prompts scroll should come down otherwise for normal agent responses scroll should remain in manual position.

jdneo: Exactly!

raghucssit:
I have fixed the problem without changing any existing behavior.
Also I have added a preference for user if they don't want to get notified about any prompt dialog, warning and error they can turn of this auto scroll functionality at preference page `Window-->Preferences..-->GitHub Copilot-->General-->Automatically scroll to new confirmation dialogs/prompts in Chat View`.

image


**What is the fix ?**
All these prompts in question are new widgets on scrolled composite. Whenever they are created we ask scrolled composite to show the newly created windget.

**How to test ?
For prompts and warning dialog about too many tool request:**
1. Set `Agent Max Requests `to `2 `at `Windows-->Preferences..-->GitHub Copilot-->Chat`
2. Then use this prompt: `I want to reproduce error "Oops, maximum tool attempts reached. You can type "continue" to proceed or rephrase your request." . please run in terminal for 10 seconds each for 3 times. after each command ran print/response some sample 30 lines.`
3. This will first ask you to run terminal tool so the prompt dialog appears. Now after first command is ran it prints some lines so cursor may be on top, or you have 10 seconds time you can scroll up and read something for second terminal tool run you will be scrolled down. Before third command runs you should get warning maxiumum tool attempts reached. Again you should be scrolled down to show this warning.

**For Error Banner:**
It is difficult to reproduce. But used some work around like below.
1. You can launch eclipse run time normally and start doing any work in Chat View.
2. Meanwhile go to command prompt and kill copilot agent. 
    search the process: `ps aux | grep -i "copilot-agent"`
    kill the process: `kill -9 <PID from above search>`
3. Now return to eclipse and initiate any request at prompt. You should get below dialog.
error_banner

4. For better checking keep a break point at the beginning of `com.microsoft.copilot.eclipse.ui.chat.ChatContentViewer.renderErrorMessage(String) `and you can scroll to the top in the Chat View.

raghucssit:
@jdneo Apologize for the long delay, I was busy at something else.
I have fixed the issue without affecting any existing behavior and also added preference for people to turn off if they don't want.
Build has not run. Please trigger the build if there are any build issues I will fix them.
I have added detailed steps to test the new fix in the above comment.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/preferences/messages.properties
```diff
@@ -43,6 +43,9 @@ preferences_page_enable_strict_ssl= Enable Strict SSL
 preferences_page_whats_new_settings= What's New
 preferences_page_enable_whats_new= Always display the 'What's New' dialog after GitHub Copilot updates.
 preferences_page_enable_whats_new_tooltip= Enable this to automatically display the "What's New" page whenever GitHub Copilot is updated.
+preferences_page_auto_scroll_group= Chat
+preferences_page_auto_scroll_to_new_dialog= Automatically scroll to new confirmation dialogs/prompts in Chat View.
+preferences_page_auto_scroll_to_new_dialog_tooltip= When enabled, the chat view will automatically scroll to show any confirmation dialogs or prompts.
```
jdneo: Hi @raghucssit

Thank you for updating the pr.

My personal feeling is that we should always keep this as true and no need to expose it as a preference. Any reason you introduce it?

raghucssit: I introduced it because to support some user preference.. For example some user wants to read some previous responses from copilot and they manually check prompts..
But I agree with you, No need to introduce it right away.. may be every user like the new fix. If any user asks for it I can introduce it later.
So I will cleanly remove preference related change.

raghucssit: I have pushed the suggested changes. Please initiate build, If everything is good we can merge it.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatContentViewer.java
```diff
@@ -419,9 +421,21 @@ public BaseTurnWidget getTurnWidget(String turnId) {
...
+    SwtUtils.invokeOnDisplayThreadAsync(() -> {
+      if (warnWidget != null && !warnWidget.isDisposed()) {
+        showControl(warnWidget);
+        scrolledToWidget = true;
+      }
+
+    }, this.getParent());
     refreshScrollerLayout();
-    scrollToLatestUserTurn();
+    if (!scrolledToWidget) {
```
jdneo: A little bit confused here.

`scrolledToWidget` is set to `true` in an async call, which means, when code is executed to here, highly likely the `scrolledToWidget` is still `false`.

What will happen if we always scroll to warn/error widget after it is created and not call `scrollToLatestUserTurn()` after that?

raghucssit: `scrollToLatestUserTurn()` this call takes the scroll back to last user turn widget. That is if i have asked some question to copilot. This behavior also not good because there is no benefit to scroll there. I wanted not to change existing behavior that's why i cam e up with this work around.
In my testing everything was working fine because async always executed before we access `scrolledToWidget`. 
But my proposal is to remove call to `scrollToLatestUserTurn()` is better.  Because itreally does not have have any benefit. 
If you agree on my proposal i will fix this.

raghucssit: @jdneo Please confirm this behavior change looks ok ? From my POV it is good to implement my proposal i don't see any benefits of old behavior.

jdneo: @raghucssit Sorry for the late response. I was working on sth else recently. Will check this ASAP.

jdneo: > But my proposal is to remove call to scrollToLatestUserTurn() is better. Because itreally does not have have any benefit.

I have the same feeling as yours. Since `showControl` already reveals the component that needs to pay attention, another `scrollToLatestUserTurn()` is not necessary.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/BaseTurnWidget.java
```diff
@@ -666,6 +669,20 @@ public CompletableFuture<LanguageModelToolConfirmationResult> requestToolExecuti
 
     this.getParent().requestLayout();
 
+    // Ensure the chat content viewer scrolls to show the newly created confirmation
+    // dialog. Walk up the composite hierarchy to find a ChatContentViewer
+    // and request scrolling. Use async exec because layout needs to complete first.
+    SwtUtils.invokeOnDisplayThreadAsync(() -> {
+      ChatContentViewer viewer = SwtUtils.findParentOfType(this.getParent(), ChatContentViewer.class);
+      if (viewer != null) {
+        viewer.refreshScrollerLayout();
```
jdneo: What will happen if we not call `viewer.refreshScrollerLayout();` here?

raghucssit:
`com.microsoft.copilot.eclipse.ui.chat.ChatContentViewer `has been redesigned without using scrolled composite. There is a mouse wheel listener is introduced to detect the need of scrolling. The fix which was originally working on this PR needs rework.
However with new design still prompts are not revealed. The problem still remains.
But warning banner like `"Maximum tool attempts"` is revealed properly. Now I only need to fix the tool prompts.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/BaseTurnWidget.java
raghucssit: `viewer.refreshScrollerLayout();` is present anymore after re design.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatContentViewer.java
raghucssit: `scrollToLatestUserTurn()` can be safely removed because it unnecessarily scrolls to latest user turn widget instead of stopping at newly created prompt. So i have removed call to this method.

raghucssit:
I have fixed the PR by adding new method `showControl() `basically it will mimic the old `showcontrol()` fo `scrolled composite.`

raghucssit:
@jdneo, Please check this PR. This is pending for long time. There were 2 major refactoring for `com.microsoft.copilot.eclipse.ui.chat.ChatContentViewer` sincle this PR was created. I had to do lot of refactoring to support it.

jdneo (dismissed):

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatContentViewer.java
```diff
@@ -770,6 +784,51 @@ private int topOf(Control target) {
...
+    }
+    // Content-coordinate top of the enclosing turn widget.
+    int ancestorTop = topOf(ancestor);
+    // Accumulate the local y offset by walking from target up to (but not including) ancestor.
+    int localY = 0;
+    for (Control c = target; c != ancestor; c = c.getParent()) {
+      localY += c.getLocation().y;
+    }
+    int targetTop = ancestorTop + localY;
+    int targetHeight = target.computeSize(getClientArea().width, SWT.DEFAULT).y;
```
jdneo: A nit here: Why we use getClientArea().width, not target.getSize().y?

raghucssit: Yes. We can use the target control height as it is.. I did little debugging and found out that the height computed is never different than actual height of target control.
I have pushed the fix.

jdneo (approved):

jdneo:
Thank you @raghucssit for your contribution!

Issue #62, opened 2025-12-12
Add Copilot preference for a chat's custom instructions loading


iloveeclipse:
Updated to latest 0.14.0 Copilot today, Eclipse 4.38.
GitHub Copilot	0.14.0.202512112209	com.microsoft.copilot.eclipse.feature.feature.group	GitHub Copilot

Chat view is unusable. Any request / prompt shows the error below, with different models in "Ask" mode on different files and with trivial questions like `/explain` or "what does this class?".

```
org.eclipse.lsp4e
Error
Fri Dec 12 09:21:45 CET 2025
[Error]	[chat] Error processing turn 861097da-6faf-43c8-9c8c-2b5eb97f9848 Error: maxTokens must be larger than the ellipsis length
    at t.elide (/snapshot/copilot-client/prompt/src/elidableText/elidableText.ts:118:19)
    at /snapshot/copilot-client/lib/src/conversation/prompt/conversationPromptEngine.ts:112:72
    at Array.map (<anonymous>)
    at q0.elideChatMessages (/snapshot/copilot-client/lib/src/conversation/prompt/conversationPromptEngine.ts:105:14)
    at q0.toPrompt (/snapshot/copilot-client/lib/src/conversation/prompt/conversationPromptEngine.ts:67:45)
    at processTicksAndRejections (node:internal/process/task_queues:105:5)
    at Zz.buildConversationPrompt (/snapshot/copilot-client/lib/src/conversation/turnProcessorStrategy.ts:79:16)
    at Bje.processWithModel (/snapshot/copilot-client/lib/src/conversation/turnProcessor.ts:189:36)
    at Bje.process (/snapshot/copilot-client/lib/src/conversation/turnProcessor.ts:90:13)
    at handleConversationCreateChecked (/snapshot/copilot-client/agent/src/methods/conversation/conversationCreate.ts:137:5)
```

It looks like the problem appears because we have few (~50) projects containing `.github/copilot-instructions.md` files in the workspace.

**If I make the file empty**, the problem disappears. 

Note: the Chat view was asked on a **fully unrelated** file from a different project, on which the instructions of other projects shouldn't apply at all.

**Update**

It looks like there is a max limit of characters that could be used in  `.github/copilot-instructions.md` file,  **multiplied by number of files** in git repository. 

With ~1700 characters for ~50 projects Chat view shows an error, with ~1300 it seem to work.

Reducing number of files also helps. 

So the bug is:

[...]

PR #133, opened 2026-04-29
feat: Support agent skills in agent mode and add preferences settings

ethanyhou:
1. Create SKILL.md files across directories:
```
    '.github/skills/**/SKILL.md',
    '.claude/skills/**/SKILL.md',
    '.agents/skills/**/SKILL.md',
    '~/.copilot/skills/**/SKILL.md',
    '~/.claude/skills/**/SKILL.md',
    '~/.agents/skills/**/SKILL.md'
```
2. Go to preference page to enable the skill:
   image
3. Switch to agent mode and type `/`, the skills should be rendered / hide corresponding to the preference settings.
   image
4. Skills is only enabled in agent mode, this is a CLS side behavior.

Issue #24, opened 2026-04-27
[Bug] Error when using search tool in workspaces with 20 or more projects open


mxalis:
**Environment**
- OS: Windows 11
- Eclipse Version: 2026-03
- Plugin Version: 0.16

**Describe the bug**
When Copilot is trying to search using the built in ripgrep tool in a larger-than-micro sized workspace (>= 20 projects, eg Maven or Gradle submodules) the tool fails with error messages related to task queue size exceeded:

[Error] [ripgrepProcessManager] Task queue limit reached: 20

**To Reproduce**
1. Open a workspace with at least 20 projects open
2. Assign a task to Copilot in Agent mode (task should require Copilot to search files)
3. Search files task fails and Copilot failovers to using findstr in command line, requiring user approval continuously. 

This pretty much forces you to constantly babysit Copilot and click Approve on the approval requests for using findstr

**Expected behavior**
Built in search tool should work even in workspaces with >= 20 projects

I understand the 20 queue limit is hard coded somewhere and it may be difficult to have it being adjustable. But it would be very much appreciated if at least the hard coded limit could be raised to something like 100 or 200.

iloveeclipse:
I've also encountered few times that agents started to search my entire workstation content (started searchat `/`), which makes no sense (and of course failed later while looking on all NFS mounts) if the only files needed were located inside the specific project.

I wonder if the plugin can use a wrapper around "built in" Eclipse text search engine API and told agents to use that wrapper instead of external grep? See `org.eclipse.search.core.text.TextSearchEngine` & friends. It would not be only faster but more accurate.

jdneo:
@iloveeclipse

If I use https://github.com/microsoft/copilot-for-eclipse/blob/f94e0373d7ae6cf26d91c35c92ed9394d70aa622/com.microsoft.copilot.eclipse.core/src/com/microsoft/copilot/eclipse/core/utils/WorkspaceUtils.java#L26 to find `'workspacefolders'` instead of `one IProject, one workspace folder`, will it breaks your case?

Or in other words, in your workspace setup, does each sub-module also contains its own .md?

iloveeclipse:
> Or in other words, in your workspace setup, does each sub-module also contains its own .md?

I believe the question belongs to another ticket? https://github.com/microsoft/copilot-for-eclipse/issues/62

In any case: we don't use nested projects, but all projects are inside one single and huge git repository.

jdneo:
[...]

PR #136, opened 2026-05-05
Add Copilot preference for a chat's custom instructions loading (from all projects or from referenced projects only)

travkin79:
This PR fixes issue #62 according to the comment https://github.com/microsoft/copilot-for-eclipse/issues/62#issuecomment-4343839724

The PR
* introduces a new setting in the Copilot preferences with currently two options for loading custom instructions in Copilot chats:
  * Load instructions from all projects in the Eclipse workspace (**new default, similar to previous behavior**)
  * Load only instructions from projects derived from files / folders selected in the chat window (from the resources' parent projects)
* if activated, derives projects only from given resources and passes the corresponding workspace folders to the language server (`CopilotLanguageServerConnection`), i.e.
  * if no files / folders are selected => no custom instructions are loaded at all
  * if a set of files / folders is selected in the chat window, only these resources' parent projects are checked for custom instructions. Custom instructions from other projects are ignored.
* adds some unit tests

**Preference page preview** (new group highlighted):
image


com.microsoft.copilot.eclipse.ui.jobs/bin/com/microsoft/copilot/eclipse/ui/jobs/i18n/messages.properties
```diff
@@ -0,0 +1,16 @@
+jobsView_toolTip_refreshAgentJobs=Refresh Agent Jobs
```
iloveeclipse: Unrelated change?

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/preferences/CopilotPreferenceInitializer.java
```diff
@@ -31,6 +32,8 @@ public void initializeDefaultPreferences() {
     pref.setDefault(Constants.AGENT_MAX_REQUESTS, 25);
     pref.setDefault(Constants.CUSTOM_INSTRUCTIONS_WORKSPACE_ENABLED, false);
     pref.setDefault(Constants.CUSTOM_INSTRUCTIONS_WORKSPACE, "");
+    pref.setDefault(Constants.CUSTOM_INSTRUCTIONS_CHAT_LOAD_SCOPE,
+        CustomInstructionsChatLoadScope.ALL_PROJECTS.getValue());
```
iloveeclipse: "All projects" shuld be default? I personally would prefer "REFERENCED_PROJECTS".
Opinions?

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/preferences/CustomInstructionPreferencePage.java
```diff
@@ -106,6 +119,22 @@ private void initializePreferenceValues() {
...
+
+    try {
+      initialChatCustomInstrLoadScope = CustomInstructionsChatLoadScope.fromValue(
+          getPreferenceStore().getString(Constants.CUSTOM_INSTRUCTIONS_CHAT_LOAD_SCOPE));
+    } catch (IllegalArgumentException e) {
+      CopilotCore.LOGGER.error("Failed to load custom instructions scope. Falling back to default value.", e);
+
+      // If the stored value is invalid, default to ALL_PROJECTS
+      getPreferenceStore().setValue(Constants.CUSTOM_INSTRUCTIONS_CHAT_LOAD_SCOPE,
+          CustomInstructionsChatLoadScope.ALL_PROJECTS.getValue());
```
iloveeclipse: Second time where `CustomInstructionsChatLoadScope.ALL_PROJECTS` is chosen as default. Better provide a field `CustomInstructionsChatLoadScope.DEFAULT_VALUE` and refer to this in all places.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/preferences/CustomInstructionPreferencePage.java
```diff
@@ -130,10 +161,33 @@ public boolean performOk() {
...
     return super.performOk();
   }
 
+  private CustomInstructionsChatLoadScope getSelectedCustomInstrLoadScope() {
+    int selectionIndex = chatInstrLoadScopeCombo.getSelectionIndex();
+
+    if (selectionIndex >= 0 && selectionIndex < SCOPE_ORDER.length) {
+      return SCOPE_ORDER[selectionIndex];
+    } else {
+      return CustomInstructionsChatLoadScope.ALL_PROJECTS;
```
iloveeclipse: third time default is hard coded

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/preferences/messages.properties
```diff
@@ -93,6 +93,10 @@ preferences_page_custom_instructions_project_editDialog_button_close=Close Page
 preferences_page_custom_instructions_project_editDialog_button_stay=Stay on Page
 preferences_page_custom_instructions_project_file_save_reminder_title=Save Reminder
 preferences_page_custom_instructions_project_file_save_reminder_desc=Remember to save the file to ensure the project instructions take effect.
+preferences_page_custom_instructions_chat= Chat
+preferences_page_custom_instructions_chat_load_scope_label= Load custom instructions
```
iloveeclipse: All new values have leading space, please remove. The code around seem to use both ways, but leading space looks wrong.

com.microsoft.copilot.eclipse.ui.jobs/bin/com/microsoft/copilot/eclipse/ui/jobs/i18n/messages.properties
travkin79: Yes, there is a .gitignore entry missing. I'll remove that change, of course.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/preferences/CopilotPreferenceInitializer.java
travkin79: "All projects" would be the same behavior as before the PR, but I also prefer "referenced projects". Let's see what the reviewers say.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/preferences/messages.properties
travkin79: fixed

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/preferences/CustomInstructionPreferencePage.java
travkin79: fixed

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/preferences/CustomInstructionPreferencePage.java
travkin79: fixed

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/preferences/CopilotPreferenceInitializer.java
travkin79: I changed the default to "referenced projects" and used a DEFAULT_VALUE constant that is being referenced everywhere where we need to use the default value.

com.microsoft.copilot.eclipse.ui.jobs/bin/com/microsoft/copilot/eclipse/ui/jobs/i18n/messages.properties
travkin79: fixed

travkin79:
Hi @iloveeclipse,
I'm not sure, if it's OK if the custom copilot instructions are only loaded when starting a new chat. The set of loaded custom instructions doesn't change in an open chat window when the set of files in the context changes. I think, as a user, I would expect the set of custom instructions to change with the file selection in the chat. WDYT?

iloveeclipse:
> I think, as a user, I would expect the set of custom instructions to change with the file selection in the chat. WDYT?

Correct.

travkin79:
Hi @iloveeclipse,

> > I think, as a user, I would expect the set of custom instructions to change with the file selection in the chat. WDYT?
> 
> Correct.

I've adapted the code to retrieve custom instructions from selected files on each conversation turn, i.e. each time a new prompt is sent to the LS.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/preferences/CustomInstructionPreferencePage.java
```diff
@@ -234,6 +288,31 @@ public void controlResized(org.eclipse.swt.events.ControlEvent e) {
         Messages.preferences_page_custom_instructions_project_table_note);
   }
 
+  private void createChatInstructionsGroup(Composite parent, GridLayout gl) {
```
jdneo: What about just directly putting the combo into the `Project` group? Since they are highly related.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/preferences/CustomInstructionPreferencePage.java
```diff
@@ -130,10 +161,33 @@ public boolean performOk() {
...
+      return SCOPE_ORDER[selectionIndex];
+    } else {
+      return CustomInstructionsChatLoadScope.DEFAULT_VALUE;
+    }
+  }
+
+  private static String getCustomInstrLoadScopeLabel(CustomInstructionsChatLoadScope scope) {
+    // the cases here must relate to SCOPE_ORDER array
+    return switch (scope) {
+      case ALL_PROJECTS -> Messages.preferences_page_custom_instructions_chat_load_scope_all;
```
jdneo, line 190: How about making the item label simpler, for example, just `All Projects` & `Referenced Projects`, and show the detailed text as their tooltip?

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatView.java
```diff
@@ -1051,6 +1058,27 @@ private void onSendInternal(String workDoneToken, String message, String agentSl
...
+    try {
+      scope = CustomInstructionsChatLoadScope.fromValue(chatInstrScope);
+    } catch (Exception e) {
+      CopilotCore.LOGGER.error(
+          "Failed parsing custom instructions load scope for chat preference, using default value", e);
+      scope = CustomInstructionsChatLoadScope.DEFAULT_VALUE;
+    }
+    return switch (scope) {
+      // take all projects from Eclipse workspace
+      case ALL_PROJECTS -> LSPEclipseUtils.getWorkspaceFolders();
```
ethanyhou, line 1133: As this is a UI package, you can use ` WorkspaceUtils.listWorkspaceFolders()` instead.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatView.java
```diff
@@ -1051,6 +1058,27 @@ private void onSendInternal(String workDoneToken, String message, String agentSl
     }
   }
 
+  List<WorkspaceFolder> deriveWorkspaceFolders(IFile currentFile, List<IResource> references) {
+    String chatInstrScope = CopilotUi.getPlugin().getPreferenceStore().getString(
+        Constants.CUSTOM_INSTRUCTIONS_CHAT_LOAD_SCOPE);
+    CustomInstructionsChatLoadScope scope;
+    try {
+      scope = CustomInstructionsChatLoadScope.fromValue(chatInstrScope);
+    } catch (Exception e) {
```
ethanyhou: Instead of broad catch here, how about catch `IllegalArgumentException`?

com.microsoft.copilot.eclipse.core/src/com/microsoft/copilot/eclipse/core/Constants.java
```diff
@@ -73,4 +76,52 @@ private Constants() {
...
+   * <li><b>REFERENCED_PROJECTS</b>: Load custom instructions only from parent projects of files/folders
+   *     referenced in the Copilot chat
+   * </ul>
+   */
+  public enum CustomInstructionsChatLoadScope {
+    ALL_PROJECTS(Constants.CUSTOM_INSTRUCTIONS_CHAT_LOAD_SCOPE_ALL),
+    REFERENCED_PROJECTS(Constants.CUSTOM_INSTRUCTIONS_CHAT_LOAD_SCOPE_REFERENCED);
+
+    public static final CustomInstructionsChatLoadScope DEFAULT_VALUE =
+        CustomInstructionsChatLoadScope.REFERENCED_PROJECTS;
```
ethanyhou: With `DEFAULT_VALUE = REFERENCED_PROJECTS`, existing users who didn't pin a value will lose custom-instructions loading whenever no files are attached.

How about keeping `ALL_PROJECTS` as the default?

cc @jdneo for double check.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/preferences/CustomInstructionPreferencePage.java
```diff
@@ -54,11 +59,18 @@ public class CustomInstructionPreferencePage extends FieldEditorPreferencePage i
   private BooleanFieldEditor enableWorkspaceInstrField;
   private StringFieldEditor workspaceInstrField;
   private StringFieldEditor gitCommitInstrField;
+  private Combo chatInstrLoadScopeCombo;
+
+  private static final CustomInstructionsChatLoadScope[] SCOPE_ORDER = {
+      CustomInstructionsChatLoadScope.ALL_PROJECTS,
+      CustomInstructionsChatLoadScope.REFERENCED_PROJECTS
+  };
```
ethanyhou: As `chatInstrLoadScopeCombo` being added, you may need to override the `perfoemOk` to make eclipse's Restore Defaults works for this new component.

com.microsoft.copilot.eclipse.core/src/com/microsoft/copilot/eclipse/core/Constants.java
travkin79: We were also not sure about the default value (see comment https://github.com/microsoft/copilot-for-eclipse/pull/136#discussion_r3188450252).

@iloveeclipse, since a preferences initializer is used, the default value could be overridden using a `plugin_customization.ini` for changing the default to `REFERENCED_PROJECT` just for the R&D team. Would it then be ok to stay with `ALL_PROJECTS` as default value in the copilot plugins?

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/preferences/CustomInstructionPreferencePage.java
travkin79: Yes, I think, you're right. I'll move the entry to the `Project` group.

com.microsoft.copilot.eclipse.core/src/com/microsoft/copilot/eclipse/core/Constants.java
jdneo: Yes please stay the all projects as default, then no behavior changes after user update the plugin.

The Distributors can customize via .ini

iloveeclipse: I'm fine with any option.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/preferences/CustomInstructionPreferencePage.java
travkin79: I checked that, but it seems, we can only add a single tooltip to the drop-down list, not to each entry. Besides, the combobox has enough space for longer text. What about adding an explaining tooltip to the combobox like that
image
and leaving relatively long, self-explaining drown-down list entries like that?
image

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatView.java
travkin79: I've checked the code and there is a significant difference between using `LSPEclipseUtils.getWorkspaceFolders()` and  `WorkspaceUtils.listWorkspaceFolders()`:

- `LSPEclipseUtils.getWorkspaceFolders()` takes all accessible projects from the Eclipse workspace
- `WorkspaceUtils.listWorkspaceFolders()` takes only top-level accessible projects from the Eclipse workspace, no nested projects (Maybe the method should be renamed to `listTopLevelWorkspaceFolders()` to make that clear?)

Nested projects could also have custom instructions that we should take into account here, don't we?

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/preferences/CustomInstructionPreferencePage.java
travkin79: Do you mean overriding `performDefaults()` as [suggested by copilot](https://github.com/microsoft/copilot-for-eclipse/pull/136#discussion_r3189930114)?

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/preferences/CustomInstructionPreferencePage.java
travkin79: Done

com.microsoft.copilot.eclipse.core/src/com/microsoft/copilot/eclipse/core/Constants.java
travkin79: I switched back to `ALL_PROJECTS` default.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/preferences/CustomInstructionPreferencePage.java
travkin79: I've added the `performDefaults()` method.

ethanyhou: Thanks, it was typo :(

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/preferences/CustomInstructionPreferencePage.java
jdneo: ok

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatView.java
jdneo: let's use LSPEclipseUtils.getWorkspaceFolders()

travkin79: Isn't this case similar to [this one](https://github.com/microsoft/copilot-for-eclipse/pull/133#discussion_r3199573836), @ethanyhou and @jdneo?

com.microsoft.copilot.eclipse.core/src/com/microsoft/copilot/eclipse/core/Constants.java
```diff
@@ -73,4 +76,52 @@ private Constants() {
...
 
+  /**
+   * Scope loading modes for custom instructions in GitHub Copilot chat.
+   * <ul>
+   * <li><b>ALL_PROJECTS</b>: Load custom instructions from all projects in the Eclipse workspace
+   * <li><b>REFERENCED_PROJECTS</b>: Load custom instructions only from parent projects of files/folders
+   *     referenced in the Copilot chat
+   * </ul>
+   */
+  public enum CustomInstructionsChatLoadScope {
```
jdneo: @travkin79

This comment does not belong to this PR's scope: There are a lot of different type of .md files used for chat feature, for example: prompt.md, skill.md, etc...

I'm thinking that maybe we can expand this setting to apply to all the md resources.

For this PR, keep the scope as it is. And I would asking that after this PR is merged, are you open to make another PR to expand this setting's scope?

jdneo: This class is now more complex than a simple const. Please consider extract it to `com/microsoft/copilot/eclipse/core/chat/CustomInstructionsChatLoadScope.java`

travkin79: Hi @jdneo,
Yes, that's an interesting idea and I could help with more PRs.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatView.java
jdneo: Yes they are similar. All md resources should be loaded in the same way.

For now, let's map ALL_PROJECT to LSPEclipseUtils.getWorkspaceFolders()

com.microsoft.copilot.eclipse.core/src/com/microsoft/copilot/eclipse/core/Constants.java
jdneo: Thank you! Tracked by https://github.com/microsoft/copilot-for-eclipse/issues/143

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/preferences/CustomInstructionPreferencePage.java
```diff
@@ -54,11 +61,18 @@ public class CustomInstructionPreferencePage extends FieldEditorPreferencePage i
...
```
jdneo: nit:

Not a big issue. Considering that currently the order is the same as the declared enum's order, and it is less likely that in future we need to add complex logic to hide or change the order of the declaration. I tend to directly use enum's declaration order to simplify the code.

```java
private static final CustomInstructionsChatLoadScope[] SCOPES =
    CustomInstructionsChatLoadScope.values();
```

Then:
- Arrays.asList(SCOPE_ORDER).indexOf(scope) becomes scope.ordinal().
- getSelectedCustomInstrLoadScope() indexes into SCOPES the same way.
- The setItems(...) stream just maps over SCOPES.
- The // the cases here must relate to SCOPE_ORDER array comment can go — the exhaustive arrow switch on the enum already forces every constant to have a label at compile time, which is a stronger guarantee than the array gave us.

travkin79: My main concern was that changing the order in the enum (or inserting a new entry) might cause loading wrong values from the preferences store. For that reason, I decoupled the combobox-index-to-enum-entry-mapping from the enum ordinals, paying the price with a more complex, but more robust code. In addition, some experts recommend not to use enum ordinals (see e.g. https://thefinestartist.com/effective-java/31).

According to comments in the other open issues / PRs (e.g. https://github.com/microsoft/copilot-for-eclipse/issues/24#issuecomment-4386332590), it's likely that we extend the enum soon.

Do you still prefer using the enum ordinal instead of an an independent combobox entries order? If we do, we should ensure to never change the order in the enum and to only append new entries to the end of the enum.

travkin79: Oh, I see that loading persisted values via `PreferencesUtils.getCustomInstructionsChatLoadScopeValue(IPreferenceStore, boolean)` and  `CustomInstructionsChatLoadScope.fromValue(String)` does not rely on enum ordinals at all, the code is already robust enough. So my concern about loading wrong values on changed enum ordering is invalid.

In that case, I'll give it a try.

Update: I implemented the suggested changes and I agree that it's indeed an improvement due to the fact that changing the order doesn't corrupt loading previously saved values.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatView.java
```diff
@@ -1051,6 +1058,27 @@ private void onSendInternal(String workDoneToken, String message, String agentSl
...
+      CopilotCore.LOGGER.error(
+          "Failed parsing custom instructions load scope for chat preference, using default value", e);
+      scope = CustomInstructionsChatLoadScope.DEFAULT_VALUE;
+    }
+    return switch (scope) {
+      // take all projects from Eclipse workspace
+      case ALL_PROJECTS -> LSPEclipseUtils.getWorkspaceFolders();
+
+      // take only projects from selected files/folders
+      case REFERENCED_PROJECTS -> ResourceUtils.deriveWorkspaceFoldersFrom(
```
jdneo, line 1136: One last question I'm not quite sure. If there is no attached file, should we treat it as no workspace folder or fallbacks to all projects?

iloveeclipse: Ideally no, causein our case we will run in same issue with Chat fully broken.

jdneo:
I have no more comments.

@iloveeclipse could you help validate with this patch, your project setup will work as expected?

iloveeclipse:
> could you help validate with this patch, your project setup will work as expected?

I hope to get some time next week for that.

jdneo:
@travkin79 Let's wait for the confirmation, thanks.

travkin79:
> > could you help validate with this patch, your project setup will work as expected?
> 
> I hope to get some time next week for that.

FYI @iloveeclipse: the PR #176 might help you setting up the Eclipse projects when you're going to test this PR.

iloveeclipse:
> the PR #176 might help you setting up the Eclipse projects when you're going to test this PR.

Thanks, but it would be also helpful if you could rebase your PR on top of master, I see there 3 merge conflicts now.

jdneo:
Sorry we have so many changes recently

iloveeclipse:
> Sorry we have so many changes recently

Not your fault, glad to see progress :-).

Interestingly, with 0.17 + this patch I don't see Copilot crashing anymore **on every chat** interaction with all instruction files enabled, just sometimes, with the original setup and both preference enabled for all projects / only in context.

I guess context window calculations is changed? What I see, and I believe it is related, that the Copilot reports following token consumption for both cases (same command, same model, new chat started)

I've asked 
> Which custom instructions files were used?

"Projects from context": 
`[LS:GitHub Copilot Language Server][Info]	[roundMetricsTracker] Turn token usage: 7413 prompt, 38 completion, 7397 cached (cache rate: 99.8%)`

"All projects":
`[LS:GitHub Copilot Language Server][Info]	[roundMetricsTracker] Turn token usage: 138937 prompt, 1795 completion, 138936 cached (cache rate: 100.0%)`.

One error I saw (with "all projects" enabled, one in 5 times) was: 
```
[LS:GitHub Copilot Language Server][Error]	[default] unhandledRejection: Error: Tests: either GITHUB_PAT, GITHUB_OAUTH_TOKEN, or GITHUB_OAUTH_TOKEN+VSCODE_COPILOT_CHAT_TOKEN must be set unless running from an IS_SCENARIO_AUTOMATION environment. Run "npm run get_token" to get credentials.
    at ddt.tokenProvider (/snapshot/cls-source/dist/main.js:1450:8372)
    at Object.get accessToken [as accessToken] (/snapshot/cls-source/dist/main.js:1319:1047)
    at ddt._handleAuthChangeEvent (/snapshot/cls-source/dist/main.js:788:11061)
    at processTicksAndRejections (node:internal/process/task_queues:103:5)
```

Not sure how "138937" vs "7413" turn tokens are counted, we have currently 51 projects with 51 identical instruction files, so I would have expected 51x higher consumption, but we see "only" 18x increase. "Completion tokens" counts 38 vs 1795 are closer to the 51x projects.

Anyway, it looks like the change works as expected.

travkin79:
> > the PR #176 might help you setting up the Eclipse projects when you're going to test this PR.
> 
> Thanks, but it would be also helpful if you could rebase your PR on top of master, I see there 3 merge conflicts now.

Done

jdneo:
Are there any difference regarding the agent response correctness between `Projects from context` and `All projects`?

iloveeclipse:
> Are there any difference regarding the agent response correctness between `Projects from context` and `All projects`?

Haven't verified that, since I was focused on stability first.

jdneo (approved):
LGTM, thank you for the contribution.

We can follow up on this one https://github.com/microsoft/copilot-for-eclipse/issues/143. Let's discuss if you have any questions :)

travkin79:
Great, thank you for reviewing and merging @jdneo. 👍

PR #161, opened 2026-05-09
fix - Integrate CLS session persistence and restoration for conversation history

xinyi-gong:
cls related PR: https://github.com/microsoft/copilot-language-server-internal/pull/893

## Summary
Integrate CLS (Copilot Language Server) server-side session persistence so that restored conversations can resume with full context instead of relying solely on IDE-side turn history.

## Changes
Session restoration protocol:

- Add conversationId and restoreToTurnId fields to ConversationCreateParams for CLS server-side restoration
- Add turnId field to Turn for tracking CLS-assigned turn IDs in the protocol layer
- Update CopilotLanguageServerConnection.createConversation() to accept and forward restoration parameters

Turn ID management:

- User turns are now persisted with turnId = null initially (instead of using workDoneToken as a fake turnId)
- CLS-assigned turnId is set via setUserTurnId() when onChatProgress(begin) arrives
- findOrCreateUserTurn / findOrCreateCopilotTurn use type-aware findTurn(conversation, turnId, Class<T>) to avoid collisions when user and copilot turns share the same turnId

Turn pairing in convertToTurns():

- User and copilot turns are now paired into a single Turn object (request + response) instead of being split into separate Turn objects, matching CLS expectations

Transcript directory configuration:

- Add transcriptDirectory field to CopilotAgentSettings
- Set ~/.copilot/eclipse as transcript directory on startup (matching IntelliJ's ~/.copilot/jb convention)
- Add PlatformUtils.getTranscriptDirectory() and Constants.TRANSCRIPT_SUBDIR


com.microsoft.copilot.eclipse.core/src/com/microsoft/copilot/eclipse/core/lsp/CopilotLanguageServerConnection.java
```diff
@@ -275,6 +275,18 @@ public CompletableFuture<ChatCreateResult> createConversation(String workDoneTok
...
+    return createConversation(workDoneToken, message, files, currentFile, currentSelection, turns, activeModel,
+        chatModeName, customChatModeId, todos, agentSlug, agentJobWorkspaceFolder, null, null);
+  }
+
+  /**
+   * Create a conversation with the given parameters, including optional conversationId and restoreToTurnId for session
+   * restoration.
+   */
+  public CompletableFuture<ChatCreateResult> createConversation(String workDoneToken, String message,
+      List<IResource> files, IFile currentFile, Range currentSelection, List<Turn> turns, CopilotModel activeModel,
```
jdneo, line 290: Since now we uses the CLS side persisted history, is `List<Turn> turns` still needed?

com.microsoft.copilot.eclipse.core/src/com/microsoft/copilot/eclipse/core/persistence/ConversationDataFactory.java
```diff
@@ -208,22 +207,38 @@ public List<Turn> convertToTurns(List<AbstractTurnData> turnDataList) {
...
     List<Turn> result = new ArrayList<>(snapshot.size());
+    Turn pendingTurn = null;
+
     for (AbstractTurnData turnData : snapshot) {
       if (turnData == null) {
         continue;
       }
       if (turnData instanceof UserTurnData userTurnData) {
+        // Flush any pending turn without a response
+        if (pendingTurn != null) {
```
jdneo: pendingTurn is confusing, what about `unpairedUserTurn`?

xinyi-gong: Updated.

com.microsoft.copilot.eclipse.core/src/com/microsoft/copilot/eclipse/core/lsp/CopilotLanguageServerConnection.java
xinyi-gong: Yes, `turns` is still needed. For conversations that were created before this PR (i.e., before CLS transcript persistence was enabled), they already have `conversationId` and` turnId` persisted on our side, but there's no corresponding transcript on the CLS side. In this case CLS needs the full turns array to reconstruct the conversation context. Additionally, even for new conversations, if the user deletes the transcript directory, `turns` serves as a fallback so the conversation history is not lost.

jdneo (approved):

Issue #163, opened 2026-05-09
 Subagent turns appear as separate assistant messages after restoration


xinyi-gong:
## Problem

When a subagent executes (e.g. run_subagent tool call), its progress events are persisted as independent CopilotTurnData entries in the conversation JSON. When the conversation is restored from history, these subagent turns appear as separate assistant messages instead of being nested under the parent agent turn.

## Root Cause

ChatView.onChatProgress() calls persistenceManager.cacheConversationProgress() for all progress events, including subagent events. updateConversationProgressInternal() creates a new CopilotTurnData for the subagent's turnId. Since subagent turnIds don't match any UserTurnData, they become orphaned copilot turns.

During restoration, restoreTurn() renders each CopilotTurnData as an independent assistant message, with no knowledge that it was a subagent execution.

## Reproduction Steps

1. Start a conversation in Agent mode
2. Send a message that triggers subagent execution
3. Wait for subagent to complete
4. Open chat history, switch to another conversation, then switch back
5. Observe: subagent tool calls appear as a separate assistant message below the parent turn

<img width="837" height="668" alt="Image" src="https://github.com/user-attachments/assets/5a0a2989-2497-480c-a264-1a6f5ad17b83" />

PR #166, opened 2026-05-09
fix -  subagent progress isolation, persistence, and restoration during session switching

### LINKED ISSUE #160: Subagent progress events leak into unrelated conversation UI when switching sessions
## Problem

When a subagent is executing and the user switches to a different conversation (via history), the subagent's progress events continue to be rendered in the **newly opened conversation's UI**, corrupting the display.

## Root Cause

ChatView.onChatProgress() is a global listener that does not filter progress events by conversationId. When the user switches conversations:

1. The subagent continues executing in the background (tied to the old conversation)
2. Progress events arrive with the old conversation's conversationId
3. onChatProgress blindly forwards them to the current chatContentViewer, which now displays a different conversation

## Expected Behavior

Progress events should be **ignored** (or buffered) if they don't belong to the currently displayed conversation. The check should account for both the main conversation ID and any active subagent conversation ID.

## Reproduction Steps

1. Start a conversation in Agent mode
2. Trigger a subagent execution
3. While the subagent is running, open chat history and switch to a different conversation
4. Observe: the subagent's output appears in the newly opened conversation's UI

xinyi-gong:
fix https://github.com/microsoft/copilot-for-eclipse/issues/160, https://github.com/microsoft/copilot-for-eclipse/issues/163
## Summary
Fixes subagent progress events leaking into unrelated conversations when switching sessions, adds conversation/destroy to stop CLS processing on cancel/switch, and correctly persists and restores subagent execution content inside SubagentMessageBlock cards.

## Changes
Progress event isolation
- Add isProgressForCurrentConversation() to filter report/end events by conversationId, with subagent awareness via subagentConversationId
- Auto-set subagentConversationId on first subagent event
- Add ConversationDestroyParams and conversation/destroy LSP method
onCancel() sends destroyConversation() to CLS for both main and subagent conversations, stopping server-side LLM calls
- Reset send button in onCancel() to prevent stuck cancel state

Subagent persistence
- Add parentTurnId and subagentToolCallId fields to CopilotTurnData
- Mark subagent progress with parentTurnId during cacheConversationProgress
- Track lastRunSubagentToolCallId from run_subagent tool calls and associate with subagent CopilotTurnData
- Skip subagent turns in convertToTurns() (not part of main conversation history sent to CLS)

Subagent restoration
- restoreTurn(): subagent CopilotTurnData (with parentTurnId) renders inside the parent turn's SubagentMessageBlock via restoreSubagentContent()
- handleSubagentToolCall(completed): creates SubagentMessageBlock for restoration path (no prior running status in persisted data)
- BaseTurnWidget.subagentBlocks map stores blocks by tool call ID for precise content injection
- Extract restoreCopilotTurnContent() as shared method for main and subagent turns

Test plan
- Add subagent.md with 5 test cases covering live execution, persistence, session switching, cancel, and multi-subagent scenarios

### DIFF
--- a/com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatView.java
+++ b/com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatView.java
@@ -46,6 +46,7 @@ import com.microsoft.copilot.eclipse.core.chat.ChatProgressListener;
 import com.microsoft.copilot.eclipse.core.chat.CustomChatModeManager;
 import com.microsoft.copilot.eclipse.core.events.CopilotEventConstants;
 import com.microsoft.copilot.eclipse.core.lsp.CopilotLanguageServerConnection;
+import com.microsoft.copilot.eclipse.core.lsp.protocol.AgentRound;
 import com.microsoft.copilot.eclipse.core.lsp.protocol.AgentToolCall;
 import com.microsoft.copilot.eclipse.core.lsp.protocol.ChatCreateResult;
 import com.microsoft.copilot.eclipse.core.lsp.protocol.ChatMode;
@@ -110,6 +111,7 @@ public class ChatView extends ViewPart implements ChatProgressListener, MessageL
   private boolean hasHistory = false;
   private String conversationId = "";
   private String subagentConversationId = null;
+  private String lastRunSubagentToolCallId = null;
   private ConversationState conversationState = ConversationState.NEW_CONVERSATION;
   private Set<CompletableFuture<?>> conversationFutures = new HashSet<>();
   private IEventBroker eventBroker = PlatformUI.getWorkbench().getService(IEventBroker.class);
@@ -804,6 +806,11 @@ public class ChatView extends ViewPart implements ChatProgressListener, MessageL
         });
         break;
       case report:
+        // Ignore progress events from a different conversation
+        if (!isProgressForCurrentConversation(value)) {
+          return;
+        }
+
         // Update context size donut if data is available
         ContextSizeInfo contextSize = value.getContextSize();
         if (contextSize != null) {
@@ -830,17 +837,43 @@ public class ChatView extends ViewPart implements ChatProgressListener, MessageL
           this.chatContentViewer.processTurnEvent(value);
         }
 
+        // Track run_subagent tool call ID for associating subagent turns
+        if (StringUtils.isBlank(value.getParentTurnId()) && value.getAgentRounds() != null) {
+          for (AgentRound round : value.getAgentRounds()) {
+            if (round.getToolCalls() != null) {
+              for (AgentToolCall tc : round.getToolCalls()) {
+                if ("run_subagent".equals(tc.getName())) {
+                  this.lastRunSubagentToolCallId = tc.getId();
+                }
+              }
+            }
+          }
+        }
+
         // If exiting subagent context (no parentTurnId), clear the subagent conversation ID
         if (StringUtils.isBlank(value.getParentTurnId()) && this.subagentConversationId != null) {
           this.subagentConversationId = null;
+          this.lastRunSubagentToolCallId = null;
         }
 
         // Cache conversation progress on report
         if (persistenceManager != null) {
           persistenceManager.cacheConversationProgress(this.conversationId, value);
+
+          // Set subagentToolCallId on subagent CopilotTurnData for restoration
+          if (StringUtils.isNotBlank(value.getParentTurnId())
+              && StringUtils.isNotBlank(this.lastRunSubagentToolCallId)) {
+            persistenceManager.setSubagentToolCallId(this.conversationId, value.getTurnId(),
+                this.lastRunSubagentToolCallId);
+          }
         }
         break;
       case end:
+        // Ignore progress events from a different conversation
+        if (!isProgressForCurrentConversation(value)) {
+          return;
+        }
+
         if (this.chatContentViewer != null) {
           this.chatContentViewer.processTurnEvent(value);
           this.actionBar.resetSendButton();
@@ -1052,6 +1085,18 @@ public class ChatView extends ViewPart implements ChatProgressListener, MessageL
   }
 
   /**
+   * Checks if a progress event belongs to the currently displayed conversation. Subagent events (with parentTurnId)
+   * must match the tracked subagent conversation ID. Non-subagent events must match the main conversation ID.
+   */
+  private boolean isProgressForCurrentConversation(ChatProgressValue value) {
+    String progressConversationId = value.getConversationId();
+    if (StringUtils.isNotBlank(value.getParentTurnId())) {
+      return StringUtils.equals(progressConversationId, this.subagentConversationId);
+    }
+    return StringUtils.equals(progressConversationId, this.conversationId);
+  }
+
+  /**
    * Align with @Workspace of vscode, because we are actually indexing the whole workspace, not a single project.
    * (@Project is only for IntelliJ.)
    *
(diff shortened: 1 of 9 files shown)


com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatView.java
```diff
@@ -1128,8 +1173,23 @@ private void clearCurrentConversation() {
 
   @Override
   public void onCancel() {
+    // Destroy the conversation on the server to stop any in-progress processing
```
jdneo: `destroy` does not means to remove this session from persistence, right?

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatView.java
```diff
@@ -830,17 +837,43 @@ public void onChatProgress(ChatProgressValue value) {
...
           this.subagentConversationId = null;
+          this.lastRunSubagentToolCallId = null;
         }
 
         // Cache conversation progress on report
         if (persistenceManager != null) {
-          persistenceManager.cacheConversationProgress(this.conversationId, value);
+          // Chain setSubagentToolCallId after cacheConversationProgress to avoid race condition
+          // where the subagent CopilotTurnData hasn't been created yet
+          final String subagentToolCallId = this.lastRunSubagentToolCallId;
```
jdneo, line 862: **🐛 Race condition: `this.conversationId` is captured by reference, not snapshotted**

The local `subagentToolCallId` is correctly snapshotted into a `final` variable before the async chain, but `this.conversationId` is **not** — it's dereferenced from `this` inside the `thenRun` continuation, which runs on a ForkJoinPool worker after `cacheConversationProgress` completes.

This is the exact scenario this PR is trying to fix (subagent state leaking across sessions). Consider this sequence:

1. A `report` event for conversation **A** arrives. The outer call correctly passes `this.conversationId = "A"` into `cacheConversationProgress(...)`, which creates/updates the subagent `CopilotTurnData` under **A**.
2. Before the `thenRun` continuation runs, the user switches sessions (or starts a new chat) on the UI thread, mutating `this.conversationId` to `"B"` (or `""`).
3. The continuation finally fires and calls:
   ```java
   persistenceManager.setSubagentToolCallId(this.conversationId, value.getTurnId(), subagentToolCallId);
   //                                       ^^^^^^^^^^^^^^^^^^^^^ now "B"
   ```
4. `setSubagentToolCallId` calls `getConversationFromCacheOrLoadFromDisk("B")` and `findTurn(conversationB, value.getTurnId())`. The subagent turn lives under **A**, so the lookup either:
   - returns `null` → silent no-op, `subagentToolCallId` is never persisted, and restoration falls back to the legacy “append to parent widget directly” path; or
   - returns a **different** turn that happens to share the same `turnId` in conversation **B** → we stamp `subagentToolCallId` onto the wrong conversation's turn.

There's also no synchronization on `this.conversationId` between the LSP/UI thread that mutates it and the ForkJoinPool worker that reads it, so even visibility across threads is not guaranteed by the JMM.

### Suggested fix

Snapshot `conversationId` the same way `subagentToolCallId` is snapshotted:

```suggestion
        // Cache conversation progress on report
        if (persistenceManager != null) {
          // Chain setSubagentToolCallId after cacheConversationProgress to avoid race condition
          // where the subagent CopilotTurnData hasn't been created yet.
          // Snapshot both conversationId and subagentToolCallId to avoid reading mutated
          // instance state inside the async continuation (e.g., after a session switch).
          final String convId = this.conversationId;
          final String subagentToolCallId = this.lastRunSubagentToolCallId;
          persistenceManager.cacheConversationProgress(convId, value).thenRun(() -> {
            if (StringUtils.isNotBlank(value.getParentTurnId()) && StringUtils.isNotBlank(subagentToolCallId)) {
              persistenceManager.setSubagentToolCallId(convId, value.getTurnId(), subagentToolCallId);
            }
          });
        }
```

Optional follow-up: have `ConversationPersistenceManager.setSubagentToolCallId(...)` return its `CompletableFuture<Void>` instead of being fire-and-forget, so callers can actually `thenCompose` on it rather than relying on the writeLock + idempotency check (`turn.getSubagentToolCallId() == null`) for ordering.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatView.java
```diff
@@ -830,17 +837,43 @@ public void onChatProgress(ChatProgressValue value) {
...
+          persistenceManager.cacheConversationProgress(this.conversationId, value).thenRun(() -> {
+            if (StringUtils.isNotBlank(value.getParentTurnId()) && StringUtils.isNotBlank(subagentToolCallId)) {
+              persistenceManager.setSubagentToolCallId(this.conversationId, value.getTurnId(), subagentToolCallId);
+            }
+          });
```
jdneo: **⚠️ The "chain" doesn't actually chain — `setSubagentToolCallId` is fire-and-forget**

The comment above this block says:
> Chain setSubagentToolCallId after cacheConversationProgress to avoid race condition where the subagent CopilotTurnData hasn't been created yet

But `setSubagentToolCallId` is declared `void` and internally schedules its own `CompletableFuture.runAsync(...)` whose returned future is discarded:

```java
// ConversationPersistenceManager.java
public void setSubagentToolCallId(String conversationId, String subagentTurnId, String toolCallId) {
  if (toolCallId == null || subagentTurnId == null) {
    return;
  }
  CompletableFuture.runAsync(() -> {       // ← fire-and-forget, future is dropped
    lock.writeLock().lock();
    try {
      ...
    } finally {
      lock.writeLock().unlock();
    }
  });
}
```

So what `thenRun` actually guarantees here is only that the **scheduling** of `setSubagentToolCallId` happens after `cacheConversationProgress` finishes — not that the *write* it performs happens after. Concretely:

- `cacheConversationProgress` for `report` event N completes → write lock released.
- `thenRun` fires and calls `setSubagentToolCallId(...)`, which submits **task X** to the ForkJoinPool.
- Before task X is picked up, `cacheConversationProgress` for `report` event N+1 is submitted as **task Y** and may grab the writeLock first.

It happens to work today only because:
1. Both tasks contend on the same `writeLock` (mutual exclusion).
2. `setSubagentToolCallId` is idempotent thanks to the `turn.getSubagentToolCallId() == null` guard.
3. `cacheConversationProgress` doesn't touch the `subagentToolCallId` field.

If any of those three properties change later (e.g., `cacheConversationProgress` starts writing `subagentToolCallId`, or someone removes the null-guard, or the lock is replaced), the ordering will silently matter and we'll get hard-to-reproduce restoration bugs.

### Suggested fix

Have `setSubagentToolCallId` return the `CompletableFuture<Void>` so callers can really chain on it:

```java
// ConversationPersistenceManager.java
public CompletableFuture<Void> setSubagentToolCallId(String conversationId, String subagentTurnId,
    String toolCallId) {
  if (toolCallId == null || subagentTurnId == null) {
    return CompletableFuture.completedFuture(null);
  }
  return CompletableFuture.runAsync(() -> {
    lock.writeLock().lock();
    try {
      ConversationData conversation = getConversationFromCacheOrLoadFromDisk(conversationId);
      if (conversation == null) {
        return;
      }
      AbstractTurnData turnData = findTurn(conversation, subagentTurnId);
      if (turnData instanceof CopilotTurnData turn && turn.getSubagentToolCallId() == null) {
        turn.setSubagentToolCallId(toolCallId);
      }
    } catch (IOException e) {
      CopilotCore.LOGGER.error("Failed to set subagent tool call ID: " + conversationId, e);
    } finally {
      lock.writeLock().unlock();
    }
  });
}
```

Then on the caller side use `thenCompose` so the returned future actually represents completion of both stages:

```java
// ChatView.java
if (persistenceManager != null) {
  final String convId = this.conversationId;                       // see issue #1
  final String subagentToolCallId = this.lastRunSubagentToolCallId;
  persistenceManager.cacheConversationProgress(convId, value)
      .thenCompose(v -> {
        if (StringUtils.isNotBlank(value.getParentTurnId())
            && StringUtils.isNotBlank(subagentToolCallId)) {
          return persistenceManager.setSubagentToolCallId(convId, value.getTurnId(), subagentToolCallId);
        }
        return CompletableFuture.completedFuture(null);
      });
}
```

This gives us a real happens-before chain instead of relying on the writeLock + idempotency guard as an implicit ordering mechanism.

xinyi-gong: Updated. Thanks for the detailed analysis!

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatView.java
xinyi-gong: Updated.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatView.java
xinyi-gong: yes, destroy will only cancel all the on-going requests from cls side. update the comment to reflect the real behavior.

jdneo (approved):
LGTM

PR #176, opened 2026-05-12
Extend CONTRIBUTING.md and adapt some Eclipse project settings to simplify getting started

travkin79:
This PR
- explains in more details how to start contributing, e.g. how to import Eclipse projects and how to launch the plug-ins
- adds `.project` files and Eclipse settings to simplify Eclipse project import for new developers
- Explains how to configure the Checkstyle Eclipse plug-in to work with this git repo

### DIFF
--- a/com.microsoft.copilot.eclipse.branding/.project
+++ b/com.microsoft.copilot.eclipse.branding/.project
@@ -1,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
+	<name>com.microsoft.copilot.eclipse.branding</name>
+	<comment></comment>
+	<projects>
+	</projects>
+	<buildSpec>
+		<buildCommand>
+			<name>org.eclipse.pde.ManifestBuilder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
+		<buildCommand>
+			<name>org.eclipse.pde.SchemaBuilder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
+		<buildCommand>
+			<name>org.eclipse.m2e.core.maven2Builder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
+	</buildSpec>
+	<natures>
+		<nature>org.eclipse.m2e.core.maven2Nature</nature>
+		<nature>org.eclipse.pde.PluginNature</nature>
+	</natures>
+</projectDescription>
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -58,9 +58,7 @@ The installable P2 repository is generated in `com.microsoft.copilot.eclipse.rep
    * Select *File > Import... > General > Existing projects into Workspace*,
      select the root directory of the copilot-for-eclipse git repo,
      activate the check box *Search for nested projects*, and finish the wizard.
-   * Do import additional projects using the import wizard for Maven project (without a `.project` file):
-     Select *File > Import... > Maven > Existing Maven Projects*,
-   * Do also import the agent bundle for your OS (e.g., `com.microsoft.copilot.eclipse.core.agent.win-x64`)
+   * Do also import the agent bundle for your OS (e.g., `com.microsoft.copilot.eclipse.core.agent.win32`)
      after building the project with npm and maven or import all OS-specific agent bundles.
 2. Activate one of the target platforms, i.e. open one of the target definition files and select `Set As Active Target Platform`.
    * target-terminal.target (Eclipse 4.37+)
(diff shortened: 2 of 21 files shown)


travkin79:
Hello @jdneo,
I have some suggestions how to simplify the first steps for new contributors. Could you please take a look at this PR? Do the suggestions make sense for you? Is it ok to have them in one single PR?

Besides, I had some questions left:
- Which of the two target definition files are you usually using for development? I took `target-tm-terminal.target`
- It seems, the launch config `launch/plugin_debug_configuration.launch` is not up-to-date or incomplete. For example, I didn't see any agent bundles selected. Are you usually using some other launch config? I guess, there should be different launch configs, one for each OS or one config explicitly for a particular OS.

jdneo:
Thank you, sorry for the confusion we have. I guess we do have some gaps here for the contribution.

> Which of the two target definition files are you usually using for development? I took target-tm-terminal.target

This is a good point. TBH, I usually just using the running platform to makes sure the launched debug session is similar to a real eclipse instance. But when it comes to debug a test case, I will switch to target-terminal.target because I use a newer Eclipse version. (There is a bundle namespace change between old eclipse and new eclipse for the terminal bundle, that's why we have two of them.)

But I admit that this setup is wired, I'm just not able to find some time organize them systematically.


> It seems, the launch config launch/plugin_debug_configuration.launch is not up-to-date or incomplete. For example, I didn't see any agent bundles selected. Are you usually using some other launch config? I guess, there should be different launch configs, one for each OS or one config explicitly for a particular OS.

Because I usually use running platform as the target platform, so when I launch a debug session, I enabled all the plugins. Maybe this is not the best practice? Any suggestions?

travkin79:
Hi @jdneo,

> ... sorry for the confusion we have.

No problem.

Concerning the target platform definitions: I think, developers not familiar with copilot-for-eclipse should preferably use one of the target definition files. I tried `target-terminal.target` at first, but the launch config seems to be prepared for the older Eclipse version, so I finally took `target-tm-terminal.target` for development. I think, TM4E project applies a reasonable approach for their target definitions that could also fit to copilot-for-eclipse. [TM4E has](https://github.com/eclipse-tm4e/tm4e/tree/main/target-platforms)
- **latest** target platform definition for the latest Eclipse release
- **oldest** target platform definition for the oldest supported Eclipse release version
- **staging** target platform definition for the latest unreleased Eclipse version

The target platforms are also [used in their CI/CD workflows](https://github.com/eclipse-tm4e/tm4e/blob/main/.github/workflows/build.yml).

> Because I usually use running platform as the target platform, so when I launch a debug session, I enabled all the plugins. Maybe this is not the best practice? Any suggestions?

I'd prefer using one of the target platform definition files instead. Otherwise, you cannot be sure that your code would compile or run with the target platforms, too. But it can help to also check the latest Eclipse version, too. TM4E does that with their latest target definition.

Usually, I'm using different launch configs for different OSs (or only for one OS) if there are platform-dependent bundles. I usually do not start any test bundles in my launch configs and I either select all or only a small sub-set of required plug-ins from the target platform (I select the required plug-ins from the Eclipse workspace and then add required plug-ins from the target platform using the "Select Required" button. Sometimes I also add some plug-ins explicitly, e.g. for Eclipse themes). In the first case I have a full Eclipse IDE, in the second case I have a lightweight IDE with short launch times. Both cases work well for me and both have their pros and cons. I'm not sure if my approach is best practice. In any case, maintaining multiple launch configs quickly becomes tedious, since developers are usually using exactly one OS and thus are maintaining only the one launch config that they are using themselves.

com.microsoft.copilot.eclipse.branding/.project
```diff
@@ -0,0 +1,40 @@
...
+			<name>org.eclipse.pde.SchemaBuilder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
+		<buildCommand>
+			<name>org.eclipse.m2e.core.maven2Builder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
+		<buildCommand>
```
jdneo: I assume there is no need to enable checkstyle for those non-code bundle. Any special reason for adding it?

also same for the java builder

travkin79: Since Copilot suggested the change, I checked the projects and the `checkstyle.xml`. The only reason for adding the checkstyle nature and builder is for checking `pom.xml`, `feature.xml`, `plugin.xml` (not existing in the modified projects), `category.xml` for checking trailing whitespaces and spaces vs. tabs. But except of pom.xml most other xml files are usually edited with special editors. Do you prefer removing it?

Concerning Java nature and builder and related `.classpath` files, you're right. Since there is no Java code, we can remove these. For some reason the Eclipse IDE added that nature when performing the Maven project import. I'll remove the Java nature for non-code projects.

jdneo: I suggest to remove them both.

If I remember correctly, so far we do not have xml rules configured for checkstyle

travkin79: I removed both.

In addition, I adapted the output folders in existing `.classpath` files to `target/classes` (it was `bin` in some projects)

jdneo (approved):
LGTM, thank you for your contribution.

PR #182, opened 2026-05-13
Fix ui deadlock

### LINKED ISSUE #179: UI freeze: deadlock between main thread and LSP listener on quota fallback
## Bug Description

The Eclipse UI freezes permanently when a 402 quota error triggers the fallback model flow. A deadlock occurs between the **main (UI) thread** and the **LSP message listener thread**.

### Thread 1: `"main"` (UI thread) — WAITING

```
CompletableFuture.get()                          ← blocks UI thread indefinitely
  ChatBaseService.getPersistentFilePath(:125)     ← lsConnection.persistence().get()
  ChatBaseService.persistUserPreference(:86)      ← synchronized method
  ModelService.setActiveModel(:351)
  ModelService.setFallBackModelAsActiveModel(:390)
  ChatContentViewer.lambda$2(:240)                ← SWT async runnable on UI thread
  SWT Display.readAndDispatch event loop
```

### Thread 2: `"LS-...#listener-0"` (LSP listener) — WAITING

```
Display.syncExec()                               ← blocks waiting for UI thread
  SwtUtils.invokeOnDisplayThread(:81)
  ChatContentViewer.processTurnEvent(:171)
  ChatView.onChatProgress(:881)
  CopilotLanguageClient.notifyProgress(:374)
  LSP4J StreamMessageProducer.listen              ← single-threaded message reader
```

## What Happens

1. User hits a **402 quota error** → `ChatContentViewer` (line 238–240) calls `setFallBackModelAsActiveModel()` **on the UI thread**
2. That calls `persistUserPreference()` → `getPersistentFilePath()` → **`CompletableFuture.get()`** — blocking the UI thread waiting for the LSP server to respond
3. Meanwhile, the Copilot language server sends a **chat progress notification**
4. The LSP listener thread tries to deliver it via **`Display.syncExec()`** — which needs the UI thread to dispatch
5. The UI thread is blocked waiting on the LSP response, and the **LSP listener is the only thread reading server messages** — so the persistence response can never arrive

**Result**: Classic circular wait → permanent UI freeze.

## Root Cause

`ChatBaseService.getPersistentFilePath()` (line 125) calls `this.lsConnection.persistence().get()` — an **unbounded blocking call** — inside a `synchronized` method that executes on the UI thread. This violates the Eclipse threading rule: _never block the UI thread on async results that depend on the same event loop_.

## Suggested Fixes

1. **Primary**: Don't call `.get()` on the UI thread in `getPersistentFilePath()`. Cache the persistence path eagerly at startup or use `.thenAccept()` to persist asynchronously.
2. **Secondary**: Change `processTurnEvent()` from `syncExec` to `asyncExec` where possible, so the LSP listener thread isn't blocked by the UI thread.
3. **Safety net**: At minimum, add a timeout — `.get(5, TimeUnit.SECONDS)` — to prevent a permanent freeze.

## Reproduction

Trigger a 402 quota exceeded error while a chat response is actively streaming (so the LSP listener thread is busy delivering progress notifications via `syncExec`).

## Environment

- macOS (aarch64)
- Java 21.0.1+12-LTS
- Eclipse 2025-12 (4.37+)

xinyi-gong:
fix https://github.com/microsoft/copilot-for-eclipse/issues/179
## Problem

When a 402 quota error occurs during chat, the UI freezes permanently due to a deadlock between the main thread and the LSP listener thread.

**Deadlock chain:**
1. LSP listener thread receives chat progress → calls `processTurnEvent()` → blocks in `Display.syncExec()` waiting for UI thread
2. UI thread handles 402 → calls `setFallBackModelAsActiveModel()` → `persistUserPreference()` → `persistence().get()` blocks waiting for LSP response
3. LSP listener thread is the only thread that reads server messages, but it's blocked in `syncExec` → the persistence response can never arrive
4. Both threads wait on each other indefinitely → permanent UI freeze

### `ModelService.java`
- `persistUserPreference()` is now called asynchronously via `CompletableFuture.runAsync()` in `setActiveModel()`, breaking the deadlock chain by keeping the UI thread unblocked.

### DIFF
--- a/com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatContentViewer.java
+++ b/com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatContentViewer.java
@@ -217,11 +217,17 @@ public class ChatContentViewer extends ScrolledComposite {
         errMsg = Messages.chat_model_unsupported_message;
       }
       if (StringUtils.isNotEmpty(errMsg)) {
-        // TODO: remove this error message replacement if statement when the CLS side warn message is aligned.
-        if (value.getCode() == 402) {
+        // Check once whether we're already on the fallback model
+        CopilotModel fallback = this.serviceManager.getModelService().getFallbackModel();
+        CopilotModel current = this.serviceManager.getModelService().getActiveModel();
+        boolean alreadyOnFallback = fallback != null && current != null
+            && fallback.getModelName().equals(current.getModelName());
+
+        // Only replace the error message with "switched to fallback" when we're NOT already
+        // on the fallback model. When already on fallback, show the original CLS message.
+        if (value.getCode() == 402 && !alreadyOnFallback) {
           CopilotPlan userPlan = this.serviceManager.getAuthStatusManager().getQuotaStatus().copilotPlan();
-          CopilotModel fallbackModel = this.serviceManager.getModelService().getFallbackModel();
-          String fallbackModelName = fallbackModel != null ? fallbackModel.getModelName()
+          String fallbackModelName = fallback != null ? fallback.getModelName()
               : Messages.chat_noQuotaView_fallbackModel;
 
           if (MenuUtils.isCfiPlan(userPlan)) {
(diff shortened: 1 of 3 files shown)


eclipse-proplus (approved):
LGTM

ethanyhou (approved):
Verified and LGTM, thanks!

PR #191, opened 2026-05-14
[TBB] Thinking block UI improvements: Label title, collapse timing, height cap, and section parsing fix

### LINKED ISSUE #178: Investigate whether to use MarkupViewer or Label to render a thinking block title
See: https://github.com/microsoft/copilot-for-eclipse/pull/157#discussion_r3224096571

xinyi-gong:
### Use Label instead of ChatMarkupViewer for thinking block header title
- Replace `ChatMarkupViewer` with a lightweight SWT `Label` for the ThinkingBlock header title ("Thinking...", "Thinking completed", etc.)
- The header title is plain text with no markdown formatting, so `ChatMarkupViewer` (which initializes a full Markdown parser, hyperlink detectors, and CSS stylesheets) is unnecessary overhead
- Register the Label with `ChatFontService` to keep chat font synchronization
- Add `SWT.WRAP` and `updateTitleWidthHint()` to handle long titles correctly on resize

image

Resolves #178

### Fix title generation fallback showing "cancelled" instead of "completed"
- When `thinking/generateTitle` returns empty or fails, the thinking block previously showed "Thinking cancelled" — but the thinking itself completed successfully, only the title generation failed
- Changed fallback to `showCompleted("Thinking completed")` instead of `showCancelled()`
- Added `.exceptionally()` handler to catch request errors, preventing the block from getting stuck in SEALED state

Before: 
image

After:
image

### Collapse thinking block immediately when thinking ends
Previously, the thinking block stayed expanded until the thinking title generation response arrived, which could take several seconds. During this gap the block showed "Thinking completed" but remained open, which felt unresponsive.

Now the block collapses as soon as thinking output stops.

### Cap thinking body height during streaming with auto-scroll
During streaming, the thinking block body is now wrapped in a ScrolledComposite with a max height of 180px. This prevents long thinking output from pushing the chat view unboundedly downward.

- Body height grows dynamically with content up to the cap, avoiding initial blank space
- Auto-scrolls to bottom as new content arrives
- User can scroll up to read earlier content; any scroll interaction disables auto-scroll
- Auto-scroll re-enables when the user scrolls back to the bottom
- On seal/cancel, the ScrolledComposite is unwrapped and body becomes a direct child, so post-finalization expand shows full content without scrollbar or scroll-wheel interception
image

### Fix ghost section from inline bold at streaming fragment boundary
Fixed a bug where inline bold text (e.g. combined with **The Ship of Theseus** and its...) could appear as a spurious section title with a bullet point at the end of the thinking block.
image

Root cause: TITLE_PATTERN required **Title** to be followed by \n or end-of-string ($), but did not require it to be at the start of a line. When a streaming fragment happened to end right after **The Ship of Theseus**, the $ anchor matched the buffer end, creating a ghost ThinkingSection. Subsequent fragments appended more text, making the match disappear — but the widget was never cleaned up.

Fix: Added (?:^|\n) anchor to require **Title** to appear at line start or string start. Added unit tests covering inline bold, mixed inline+standalone, and title-at-start scenarios.

### DIFF
--- a/com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ThinkingBlock.java
+++ b/com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ThinkingBlock.java
@@ -34,7 +34,8 @@ import com.microsoft.copilot.eclipse.ui.utils.UiUtils;
  */
 public class ThinkingBlock extends Composite {
   private static final String SECONDARY_TEXT_CSS_CLASS = "text-secondary";
-  private static final Pattern TITLE_PATTERN = Pattern.compile("\\*\\*([^*\\r\\n]+?)\\*\\*(?=\\r?\\n|$)");
+  private static final Pattern TITLE_PATTERN =
+      Pattern.compile("(?:^|\\n)\\*\\*([^*\\r\\n]+?)\\*\\*(?=\\r?\\n|$)");
 
   private Composite header;
   private Label iconLabel;
--- a/com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ThinkingBlock.java
+++ b/com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ThinkingBlock.java
@@ -12,6 +12,7 @@ import java.util.regex.Pattern;
 import org.apache.commons.lang3.StringUtils;
 import org.eclipse.e4.ui.services.IStylingEngine;
 import org.eclipse.swt.SWT;
+import org.eclipse.swt.custom.ScrolledComposite;
 import org.eclipse.swt.events.MouseAdapter;
 import org.eclipse.swt.events.MouseEvent;
 import org.eclipse.swt.graphics.Cursor;
@@ -37,16 +38,22 @@ public class ThinkingBlock extends Composite {
   private static final Pattern TITLE_PATTERN =
       Pattern.compile("(?:^|\\n)\\*\\*([^*\\r\\n]+?)\\*\\*(?=\\r?\\n|$)");
 
+  private static final int STREAMING_MAX_HEIGHT = 180;
+
   private Composite header;
   private Label iconLabel;
   private Label titleLabel;
   private Label chevronLabel;
 
-  /** Body container holding one {@link ThinkingSection} per parsed section. Hidden when collapsed. */
+  /** Scrollable wrapper around {@link #body}; used only during streaming. Disposed on finalized expand. */
+  private ScrolledComposite bodyScroller;
+  /** Body container holding one {@link ThinkingSection} per parsed section. */
   private Composite body;
   private final List<ThinkingSection> sections = new ArrayList<>();
   private final StringBuilder textBuffer = new StringBuilder();
   private boolean expanded = true;
+  /** Auto-scroll to bottom during streaming. Turned off on any user scroll interaction. */
+  private boolean autoScroll = true;
 
   /**
    * Lifecycle of the block. Transitions are always forward: STREAMING → (SEALED →)? → COMPLETED|CANCELLED.
@@ -123,6 +130,7 @@ public class ThinkingBlock extends Composite {
     }
     setTitleText(Messages.thinking_cancelledTitle);
     setExpanded(false);
+    unwrapBodyFromScroller();
     state = State.CANCELLED;
   }
 
@@ -139,6 +147,7 @@ public class ThinkingBlock extends Composite {
     stopSpinner();
     setTitleText(Messages.thinking_completedTitle);
     setExpanded(false);
+    unwrapBodyFromScroller();
   }
 
   /** True only while new thinking stream fragments should still be appended to this block. */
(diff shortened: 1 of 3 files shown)


com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ThinkingTurnWidget.java
```diff
@@ -82,7 +82,7 @@ public void sealThinking() {
     }
     CopilotLanguageServerConnection ls = CopilotCore.getPlugin().getCopilotLanguageServer();
     if (ls == null) {
-      target.showCancelled();
+      target.showCompleted(Messages.thinking_completedTitle);
```
jdneo, line 86: Since now completed will not collapse and stop spinner, do we need to call seal first?

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ThinkingBlock.java
```diff
@@ -272,9 +291,41 @@ private void refreshBody() {
...
+  private void updateScrollerDuringStreaming() {
+    if (bodyScroller == null || bodyScroller.isDisposed()) {
+      return;
+    }
+    int contentWidth = bodyScroller.getClientArea().width;
+    if (contentWidth <= 0) {
+      contentWidth = SWT.DEFAULT;
+    }
+    int contentHeight = body.computeSize(contentWidth, SWT.DEFAULT).y;
+    bodyScroller.setMinSize(body.computeSize(contentWidth, SWT.DEFAULT));
```
ethanyhou: `body.computeSize(contentWidth, SWT.DEFAULT)` can be extracted.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ThinkingTurnWidget.java
xinyi-gong: Good catch. Updated.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ThinkingBlock.java
```diff
@@ -222,20 +229,32 @@ public void mouseUp(MouseEvent e) {
...
     bodyLayout.verticalSpacing = 6;
     body.setLayout(bodyLayout);
-    body.setLayoutData(new GridData(SWT.FILL, SWT.FILL, true, false));
+
+    bodyScroller.setContent(body);
+
+    // Any user scroll interaction disables auto-scroll unconditionally.
+    Runnable disableAutoScroll = () -> autoScroll = false;
+    bodyScroller.getVerticalBar().addListener(SWT.Selection, e -> disableAutoScroll.run());
+    body.addListener(SWT.MouseVerticalWheel, e -> disableAutoScroll.run());
```
ethanyhou, line 258: Double check if the `MouseVerticalWheel` listener needs to be disposed manually.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ThinkingBlock.java
```diff
@@ -367,17 +414,35 @@ private void toggleExpanded() {
...
+    if (bodyScroller == null || bodyScroller.isDisposed() || body == null || body.isDisposed()) {
+      return;
+    }
+    bodyScroller.setContent(null);
+    body.setParent(this);
+    GridData bodyData = new GridData(SWT.FILL, SWT.FILL, true, false);
+    bodyData.exclude = !expanded;
+    body.setLayoutData(bodyData);
+    body.setVisible(expanded);
+    body.moveBelow(bodyScroller);
```
ethanyhou: It seems like the ` body.moveBelow(bodyScroller);` can be removed.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ThinkingBlock.java
```diff
@@ -185,16 +194,14 @@ private void createHeader() {
...
-    titleText.setLayoutData(titleData);
-    titleText.setEditable(false);
-    // Strip StyledText's intrinsic margins so the text sits flush with the adjacent labels.
-    titleText.setMargins(0, 0, 0, 0);
-    UiUtils.applyCssClass(titleText, SECONDARY_TEXT_CSS_CLASS, stylingEngine);
+    titleLabel = new Label(header, SWT.LEFT | SWT.WRAP);
+    titleLabel.setLayoutData(new GridData(SWT.LEFT, SWT.CENTER, false, false));
+    UiUtils.applyCssClass(titleLabel, SECONDARY_TEXT_CSS_CLASS, stylingEngine);
+
+    var chatServiceManager = CopilotUi.getPlugin().getChatServiceManager();
```
ethanyhou: `ChatServiceManager`

ethanyhou (commented):
Overall a solid improvement. Just some nit comments. Thanks!

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ThinkingBlock.java
xinyi-gong: Updated

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ThinkingBlock.java
xinyi-gong: Checked. This will be automatically disposed by swt listener.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ThinkingBlock.java
xinyi-gong: Updated

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ThinkingBlock.java
xinyi-gong: Updated

jdneo (approved):

jdneo:
https://github.com/microsoft/copilot-for-eclipse/issues/202

PR #194, opened 2026-05-14
feat: [TBB] add model thinking effort selection (Thinking Effort Part2)

### LINKED ISSUE #204: Support selecting thinking effort for model

ethanyhou:
close #204 

Test:
1.The model with thinking ability should have thinking effort on its model selector hover popup：
image
2. After the thinking effort is selected, the thinking effort should also be shown as a suffix of the selected model displayName：
image
3. For TBB enabled accounts, the thinking effort should also be shown on the model item:
 Screenshot 2026-05-14 174600
5. For the default thinking effort, Claude should be high and other models should be Medium.

#204


jdneo:
Haven't review the code yet. There is a ux feedback. See:


https://github.com/user-attachments/assets/4671f249-0c40-422f-8b96-f53ee8642d86

You can see that if the mouse moves not fast enough and stay on the scroll bar for a short while, the detailed pane will be disposed. Making it difficult to select thinking effort.

Does windows has this issue? What about moving the detailed pane on the left.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/utils/ModelUtils.java
```diff
@@ -188,4 +191,85 @@ public static String formatTokenCount(int tokens) {
...
+   *
+   * @param model the model
+   * @return the composite key, or {@code null} if {@code model} is {@code null}
+   */
+  public static String getModelKey(CopilotModel model) {
+    if (model == null) {
+      return null;
+    }
+    return model.getProviderName() != null ? model.getProviderName() + "_" + model.getId() : model.getId();
+  }
```
jdneo: nit: Seems more suitable to be a static method of CopilotModel

com.microsoft.copilot.eclipse.core/src/com/microsoft/copilot/eclipse/core/lsp/protocol/CopilotModel.java
```diff
@@ -44,12 +44,29 @@ public String toString() {
...
+   *     (the language server only reports {@code true} when the model has more than one effort level and is hosted on
+   *     a compatible endpoint)
    */
-  public record CopilotModelCapabilitiesSupports(boolean vision) {
+  public record CopilotModelCapabilitiesSupports(boolean vision, List<String> reasoningEfforts,
+      boolean supportsReasoningEffortLevel) {
+    /**
+     * Backwards-compatible constructor used by call sites that do not provide reasoning-effort metadata.
+     */
+    public CopilotModelCapabilitiesSupports(boolean vision) {
```
jdneo: Seems no one uses it now, remove

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/services/ModelService.java
```diff
@@ -404,6 +402,108 @@ public boolean isVisionSupported() {
...
+    preference.setReasoningEffort(key, reasoningEffort);
+    CompletableFuture.runAsync(this::persistUserPreference);
+    // Rebuild bound model picker items so the row suffix (which includes the user-selected effort) reflects the
+    // new value the next time the dropdown opens. setItemGroups only affects subsequent open() calls; the
+    // currently open popup (if any) is closed explicitly by the caller after this method returns.
+    ensureRealm(this::refreshBoundModelPickers);
+    // Re-publish the active model so observers (e.g. model picker) refresh derived UI such as the picker tooltip.
+    if (Objects.equals(ModelUtils.getModelKey(getActiveModel()), key)) {
+      ensureRealm(() -> activeModelObservable.setValue(model));
+    }
```
jdneo: Can we introduce another IObservableValue for reasoning effort and driving the ui update?

Below is suggestion generated by AI:

Suggestion: replace imperative widget push with an observable in setSelectedReasoningEffort

  The setter currently mutates state, reaches into every bound picker via refreshBoundModelPickers, and republishes activeModelObservable purely as a re-render trigger. This couples ModelService to picker
  internals and overloads activeModelObservable — observers can't tell "active model changed" from "effort changed."

  Since ModelService already uses observables, adding a third for the effort state fits the existing pattern:

  private final IObservableValue<Map<String, String>> reasoningEffortObservable =
      new WritableValue<>(Map.of(), Map.class);

  The setter becomes pure state mutation:

  preference.setReasoningEffort(key, reasoningEffort);
  CompletableFuture.runAsync(this::persistUserPreference);
  ensureRealm(() -> reasoningEffortObservable.setValue(preference.getReasoningEffortSnapshot()));

  bindModelPicker subscribes and funnels into a single rebuildItems(picker) — the same method the model-list and active-model listeners call. Three triggers, one render path.

  This removes refreshBoundModelPickers and the activeModelObservable republish, keeps observable semantics honest, and makes the effort flow unit-testable without SWT.

  Not a blocker, just a cleaner shape.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/swt/DropdownButton.java
```diff
@@ -214,6 +227,13 @@ private void paintControl(GC gc) {
     }
   }
 
+  private String resolveDisplayText(DropdownItem selected) {
+    if (selectedDisplayText != null && !selectedDisplayText.isEmpty()) {
+      return selectedDisplayText;
+    }
+    return selected != null ? selected.getLabel() : "";
+  }
```
jdneo, line 224: Suggestion: move the button-face label onto DropdownItem instead of overriding it on DropdownButton

  resolveDisplayText in DropdownButton currently has to reconcile two independent state slots — selectedItemId (looked up via item label) and selectedDisplayText (an external override). They're set separately,
  can drift, and the override is silently sticky once non-blank. The picker also ends up with two render paths: buildModelDropdownItems formats the row suffix, and buildModelButtonDisplayText formats the button
  face — duplicated string-join logic over the same (model, effort) data.

  A cleaner shape is to let the selected item carry its own button-face label:

  public class DropdownItem {
    private final String label;          // shown in the dropdown row
    private final String selectedLabel;  // shown on the button when this item is selected
                                         // (defaults to label when null)
    ...
  }

  DropdownButton collapses to:

  private String resolveDisplayText(DropdownItem selected) {
    if (selected == null) return "";
    return selected.getSelectedLabel() != null ? selected.getSelectedLabel() : selected.getLabel();
  }

  ModelPickerGroupsBuilder sets selectedLabel = modelName + " - " + effort when an effort applies, and the buildModelButtonDisplayText / setSelectedDisplayText machinery goes away.

  Benefits:
  - One source of truth per item — row label and button label live on the same object.
  - setSelectedItemId becomes sufficient; no more dual-setter coordination in refreshBoundModelPickers and bindModelPicker.
  - Switching items naturally switches the button text — no "sticky override" footgun.
  - Removes ~30 lines: selectedDisplayText field, setSelectedDisplayText, buildModelButtonDisplayText, and the extra setter calls.

  Not a blocker, but it would let DropdownButton stay ignorant of why the button text differs from the row text.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/swt/ModelHoverContentProvider.java
```diff
@@ -144,6 +161,191 @@ private void addContextWindowSection(Composite parent, CopilotModelCapabilitiesL
...
+      // hover popup while the effort label stays anchored to the left.
+      descriptionLabel.setLayoutData(new GridData(SWT.RIGHT, SWT.CENTER, true, false));
+      setCssClass(descriptionLabel, POPUP_SECONDARY_TEXT_CLASS);
+    }
+
+    Cursor handCursor = parent.getDisplay().getSystemCursor(SWT.CURSOR_HAND);
+    row.setCursor(handCursor);
+    iconLabel.setCursor(handCursor);
+    optionLabel.setCursor(handCursor);
+    optionLabel.setToolTipText(displayText);
```
jdneo: nit: We can just not use the tooltip if the tooltip is the same as the displayed content

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/swt/ModelHoverContentProvider.java
```diff
@@ -144,6 +161,191 @@ private void addContextWindowSection(Composite parent, CopilotModelCapabilitiesL
...
+    if (descriptionLabel != null) {
+      descriptionLabel.setCursor(handCursor);
+      descriptionLabel.setToolTipText(description);
```
jdneo: nit: We can just not use the tooltip if the tooltip is the same as the displayed content

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/swt/ModelHoverContentProvider.java
```diff
@@ -144,6 +161,191 @@ private void addContextWindowSection(Composite parent, CopilotModelCapabilitiesL
...
+    optionsLayout.marginWidth = 0;
+    optionsLayout.marginHeight = 0;
+    optionsLayout.verticalSpacing = 2;
+    options.setLayout(optionsLayout);
+
+    populateThinkingEffortOptions(options, efforts, closeRequest);
+  }
+
+  private void populateThinkingEffortOptions(Composite options, List<String> efforts, Runnable closeRequest) {
+    for (Control child : options.getChildren()) {
```
jdneo: In current code, options is just created before this function, so it won't have any children to dispose.

Is this for defensive purpose?

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/utils/ModelUtils.java
ethanyhou: Moved

com.microsoft.copilot.eclipse.core/src/com/microsoft/copilot/eclipse/core/lsp/protocol/CopilotModel.java
ethanyhou: Removed

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/swt/ModelHoverContentProvider.java
ethanyhou: Removed

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/swt/ModelHoverContentProvider.java
ethanyhou: Removed

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/swt/ModelHoverContentProvider.java
ethanyhou: Yes, but still workable after remove. Removed

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/services/ModelService.java
ethanyhou: Updated

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/swt/DropdownButton.java
ethanyhou: Updated

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/services/ModelService.java
```diff
@@ -404,6 +402,109 @@ public boolean isVisionSupported() {
...
+   * {@link ModelUtils#resolveDefaultReasoningEffort(CopilotModel)}. Returns {@code null} for models that do not
+   * support reasoning-effort selection or for the special "auto" model.
+   *
+   * @param model the model that will receive the request
+   * @return the effort to send, or {@code null} to omit
+   */
+  public String resolveEffectiveReasoningEffort(CopilotModel model) {
+    String selected = getSelectedReasoningEffort(model);
+    if (StringUtils.isNotBlank(selected)) {
+      return selected;
```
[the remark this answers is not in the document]
ethanyhou: Updated.

ethanyhou:
> Haven't review the code yet. There is a ux feedback. See:
> 
>  Screen.Recording.2026-05-14.at.18.34.13.mov 
> You can see that if the mouse moves not fast enough and stay on the scroll bar for a short while, the detailed pane will be disposed. Making it difficult to select thinking effort.
> 
> Does windows has this issue? What about moving the detailed pane on the left.

Can repro on Windows. Fixed by checking boundary using shell.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/swt/ModelHoverContentProvider.java
```diff
@@ -144,6 +161,183 @@ private void addContextWindowSection(Composite parent, CopilotModelCapabilitiesL
...
+        return Messages.model_reasoningEffort_low_description;
+      case "medium":
+        return Messages.model_reasoningEffort_medium_description;
+      case "high":
+        return Messages.model_reasoningEffort_high_description;
+      case "xhigh":
+        return Messages.model_reasoningEffort_xhigh_description;
+      default:
+        return null;
+    }
```
jdneo: Seems like can be replaced by ModelUtils.formatReasoningEffort()?

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/services/ModelService.java
jdneo: nit: not this PR's scope.

Want to confirm that should we call .getModelKey() or .getId() here?

Currently since copilot model does not have provider name so both are fine.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/swt/ModelHoverContentProvider.java
ethanyhou: These two functions are for different purpose:

- formatReasoningEffort: format Low, Medium, High
- resolveReasoningEffortDescription: format description for the related reasoning effort.

Unified the function position and updated function name for better understanding.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/services/ModelService.java
ethanyhou: Updated to `getModelKey`

jdneo (approved):

PR #210, opened 2026-05-15
feat - Persist thinking blocks to conversation history and restore on session switch

### LINKED ISSUE #201: Persist thinking block

xinyi-gong:
resolve https://github.com/microsoft/copilot-for-eclipse/issues/201
## Change 1: Persist thinking blocks to conversation history

Persist thinking block content, state, and title so thinking blocks can be restored when switching sessions or loading conversation history.

## Data Model

- Add `ThinkingBlockData` with `id`, `content`, `title`, and `state` fields.
- Add `ThinkingBlockState` enum for `COMPLETED` and `CANCELLED`.
- Add `thinkingBlock` to `EditAgentRoundData` so thinking content is stored with the agent round it belongs to.
- Use a UI-generated thinking block ID as the stable identity for title/cancel updates instead of relying on server thinking IDs or round IDs.

## Persistence

- Accumulate thinking stream fragments directly into a thinking block attached to an edit-agent round.
- Create a placeholder edit-agent round when thinking arrives before the corresponding `AgentRound`.
- When the real `AgentRound` arrives, merge it into the placeholder by matching the UI-generated thinking block ID.
- Update thinking titles by `thinkingBlockId`.
- Update cancelled thinking state by `thinkingBlockId`.
- Keep placeholder round IDs as a synthetic sentinel (`-1`) until the real server round arrives.

## UI

- Generate a stable thinking block ID in `ThinkingBlock`.
- Add `setConversationContext(conversationId, persistTurnId)` to route persistence context to the active widget, including subagent widgets.
- `sealThinking()` uses stored context and persists generated titles by thinking block ID.
- `onChatMessageCancelled()` routes through `getActiveTurnWidget()` so subagent thinking cancellation is handled correctly.
- `ChatContentViewer` caches the active thinking block ID per turn while processing progress events, so `ChatView` can pass the correct ID to persistence without querying widget state again.
- Sync `conversationId` to `ChatContentViewer` on begin events.

## Restore

- Restore thinking blocks from persisted `EditAgentRoundData.thinkingBlock`.
- Restore main-agent thinking blocks before their corresponding round reply/tool calls.
- Restore subagent thinking blocks inside the subagent message block.
- Preserve completed/cancelled state, generated title, and full thinking content.

## Validation

- Added persistence coverage for thinking content, title updates, cancellation, and placeholder round merging.
- Added parser regression coverage for adjacent thinking section titles.
- Added manual test plans

## Change 2: Fix ThinkingBlock.parseSections crash on adjacent titles

When two bold title patterns (e.g. **Title1** immediately followed by **Title2**) appear with no body text between them, the newline-skipping logic advances the cursor past the next match's start position, causing a StringIndexOutOfBoundsException. This crashes the entire conversation restore and leaves the chat panel broken.

Fix: guard the substring call with a cursor <= matcher.start() check, producing an empty body when titles are adjacent.
Stack trace:
```
java.lang.StringIndexOutOfBoundsException: Range [45, 44) out of bounds for length 2694
	at java.base/jdk.internal.util.Preconditions$1.apply(Preconditions.java:55)
	at java.base/jdk.internal.util.Preconditions$1.apply(Preconditions.java:52)
	at java.base/jdk.internal.util.Preconditions$4.apply(Preconditions.java:213)
	at java.base/jdk.internal.util.Preconditions$4.apply(Preconditions.java:210)
	at java.base/jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:98)
	at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckFromToIndex(Preconditions.java:112)
	at java.base/jdk.internal.util.Preconditions.checkFromToIndex(Preconditions.java:349)
	at java.base/java.lang.String.checkBoundsBeginEnd(String.java:4849)
	at java.base/java.lang.String.substring(String.java:2823)
	at com.microsoft.copilot.eclipse.ui.chat.ThinkingBlock.parseSections(ThinkingBlock.java:351)
	at com.microsoft.copilot.eclipse.ui.chat.ThinkingBlock.refreshBody(ThinkingBlock.java:275)
	at com.microsoft.copilot.eclipse.ui.chat.ThinkingBlock.appendText(ThinkingBlock.java:101)
......
```
## Change 3 : Fix Thinking Stream Whitespace Handling

- Preserve whitespace-only thinking stream fragments instead of filtering them with `StringUtils.isBlank`.
- This fixes section parsing when the language server sends a newline as its own thinking fragment before the next `**Title**`.
- Use `StringUtils.isEmpty` for thinking fragments so `null` / empty strings are ignored, while meaningful markdown whitespace such as `"\n"` is preserved.
- Persist the same whitespace-only thinking fragments so restored thinking content matches the live rendered content.

Before: 
image

### DIFF
--- a/com.microsoft.copilot.eclipse.core/src/com/microsoft/copilot/eclipse/core/persistence/ConversationPersistenceManager.java
+++ b/com.microsoft.copilot.eclipse.core/src/com/microsoft/copilot/eclipse/core/persistence/ConversationPersistenceManager.java
@@ -317,7 +317,7 @@ public class ConversationPersistenceManager {
   }
 
   /**
-  * Sets the title on the thinking block with the given ID.
+   * Sets the title on the thinking block with the given ID.
    *
    * @param conversationId the conversation ID
    * @param turnId the turn ID
@@ -343,7 +343,7 @@ public class ConversationPersistenceManager {
   }
 
   /**
-  * Cancels the thinking block with the given ID.
+   * Cancels the thinking block with the given ID.
    *
    * @param conversationId the conversation ID
    * @param turnId the turn ID
@@ -394,7 +394,6 @@ public class ConversationPersistenceManager {
     }
     return null;
   }
-
 
   /**
    * Finds a turn by ID and type in the conversation.
--- a/com.microsoft.copilot.eclipse.core/src/com/microsoft/copilot/eclipse/core/persistence/ConversationPersistenceManager.java
+++ b/com.microsoft.copilot.eclipse.core/src/com/microsoft/copilot/eclipse/core/persistence/ConversationPersistenceManager.java
@@ -24,6 +24,10 @@ import com.microsoft.copilot.eclipse.core.lsp.protocol.CopilotModel;
 import com.microsoft.copilot.eclipse.core.lsp.protocol.TodoItem;
 import com.microsoft.copilot.eclipse.core.lsp.protocol.Turn;
 import com.microsoft.copilot.eclipse.core.lsp.protocol.codingagent.CodingAgentMessageRequestParams;
+import com.microsoft.copilot.eclipse.core.persistence.CopilotTurnData.EditAgentRoundData;
+import com.microsoft.copilot.eclipse.core.persistence.CopilotTurnData.ReplyData;
+import com.microsoft.copilot.eclipse.core.persistence.CopilotTurnData.ThinkingBlockData;
+import com.microsoft.copilot.eclipse.core.persistence.CopilotTurnData.ThinkingBlockState;
 import com.microsoft.copilot.eclipse.core.persistence.UserTurnData.MessageData;
 
 /**
@@ -201,12 +205,18 @@ public class ConversationPersistenceManager {
 
   /**
    * Updates a conversation with progress data and caches it only (no disk persistence).
+   *
+   * @param conversationId the conversation ID
+   * @param progress the progress value
+   * @param thinkingBlockId the UI-generated thinking block ID, when the progress belongs to a thinking round
    */
-  public CompletableFuture<Void> cacheConversationProgress(String conversationId, ChatProgressValue progress) {
+  public CompletableFuture<Void> cacheConversationProgress(String conversationId, ChatProgressValue progress,
+      String thinkingBlockId) {
     return CompletableFuture.runAsync(() -> {
       lock.writeLock().lock();
       try {
-        ConversationData conversationData = updateConversationProgressInternal(conversationId, progress);
+        ConversationData conversationData = updateConversationProgressInternal(conversationId, progress,
+            thinkingBlockId);
         conversationCache.put(conversationId, conversationData);
       } catch (Exception e) {
         CopilotCore.LOGGER.error("Failed to cache conversation progress: " + conversationId, e);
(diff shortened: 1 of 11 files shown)


com.microsoft.copilot.eclipse.core/src/com/microsoft/copilot/eclipse/core/persistence/CopilotTurnData.java
```diff
@@ -773,4 +798,102 @@ public String toString() {
...
+    public String toString() {
+      ToStringBuilder builder = new ToStringBuilder(this);
+      builder.append("id", id);
+      builder.append("content", content);
+      builder.append("title", title);
+      builder.append("state", state);
+      return builder.toString();
+    }
+  }
 }
```
ethanyhou: Restore the trailing newline

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ThinkingTurnWidget.java
```diff
@@ -81,29 +110,34 @@ public void sealThinking() {
...
-          }
-          if (resp != null && StringUtils.isNotBlank(resp.title())) {
-            target.setTitle(resp.title());
-          }
-          target.showCompleted();
-          requestLayout();
-        }, this))
+    // Chain: finalize pending thinking block first (returns the server roundId it attached to),
+    // then request title, then persist title using that exact roundId.
+    active.finalizeLastThinkingBlock(ThinkingBlockState.COMPLETED, active.persistTurnId)
```
ethanyhou: `finalizeLastThinkingBlock` here can cause race condition as the thinking report event may not cached at `ChatView.java`: persistenceManager.cacheConversationProgress(currentConversationId, value). Please take a second look.

xinyi-gong: Good catch. I agree this was race-prone.

I removed the `finalizeLastThinkingBlock` / “last round” association path. The implementation now uses a UI-generated `thinkingBlockId`: `ChatContentViewer` tracks the active thinking block ID per turn, `ChatView` passes it into persistence, and thinking fragments are accumulated directly into a placeholder round keyed by that ID. The real `AgentRound` is merged into the placeholder later, and title/cancel updates are also done by `thinkingBlockId`.

So title/final state persistence no longer depends on whether the thinking report event has already been cached.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ThinkingTurnWidget.java
```diff
@@ -115,16 +142,71 @@ public void sealThinking() {
...
-        requestLayout();
+      ThinkingTurnWidget active = getActiveTurnWidget();
+      if (active == null || active.isDisposed()) {
+        return;
+      }
+      if (active.currentBlock != null && !active.currentBlock.isDisposed()
+          && !active.currentBlock.isFinalized()) {
+        boolean cancelled = active.currentBlock.showCancelled();
+        if (cancelled) {
+          String cancelTurnId = active.persistTurnId != null ? active.persistTurnId : active.turnId;
```
jdneo: Is the fallback id safe here? Previous comments says persistTurnId and turnId are different things

com.microsoft.copilot.eclipse.core/src/com/microsoft/copilot/eclipse/core/persistence/ConversationPersistenceManager.java
```diff
@@ -299,6 +316,86 @@ private ConversationData updateConversationProgressInternal(String conversationI
...
+    for (EditAgentRoundData round : rounds) {
+      ThinkingBlockData block = round.getThinkingBlock();
+      if (block != null && thinkingBlockId.equals(block.getId())) {
+        return block;
+      }
+    }
+    return null;
+  }
+
+
```
jdneo: nit: extra blank line

xinyi-gong: Updated

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ThinkingTurnWidget.java
xinyi-gong: Good point. The fallback was not safe. I removed the fallback. Cancellation is now persisted only when `persistTurnId` is available. If it is not available, the UI still updates the thinking block to cancelled, but we avoid writing to persistence with a potentially wrong turn ID.

jdneo (approved):
LGTM

PR #240, opened 2026-05-20
fix - Persist cancelled tool call status when chat is cancelled

### LINKED ISSUE #239: [Bug] Cancelled terminal tool calling status will become ongoing again after chat history restoration
Repro:
1. Let Copilot using run in terminal tool to do some tasks.
2. Cancel the terminal tool and make sure the icon is updated to cancel icon
3. New conversation then restore the conversation
4. The cancel icon will become rotating again:
Image

xinyi-gong:
## Summary

Fixes microsoft/copilot-for-eclipse#239.

When a chat response is cancelled while an agent tool call is still running, the live UI shows the tool as cancelled, but the cached conversation can still persist that tool call with `running` status. Restoring the conversation later reads that persisted status and shows the tool call as running/spinning again.

This change marks cached `running` tool calls as `cancelled` before persisting the conversation during the chat cancel flow.

## Changes

- Add persistence handling to convert cached running tool calls to cancelled on chat cancel.
- Keep the fix scoped to the conversation persistence model, instead of changing confirmation dialog or tool skip behavior.
- Add coverage to verify only `running` tool calls are updated while completed tool calls remain unchanged.

### DIFF
--- a/com.microsoft.copilot.eclipse.core/src/com/microsoft/copilot/eclipse/core/persistence/ConversationPersistenceManager.java
+++ b/com.microsoft.copilot.eclipse.core/src/com/microsoft/copilot/eclipse/core/persistence/ConversationPersistenceManager.java
@@ -28,6 +28,7 @@ import com.microsoft.copilot.eclipse.core.persistence.CopilotTurnData.EditAgentR
 import com.microsoft.copilot.eclipse.core.persistence.CopilotTurnData.ReplyData;
 import com.microsoft.copilot.eclipse.core.persistence.CopilotTurnData.ThinkingBlockData;
 import com.microsoft.copilot.eclipse.core.persistence.CopilotTurnData.ThinkingBlockState;
+import com.microsoft.copilot.eclipse.core.persistence.CopilotTurnData.ToolCallData;
 import com.microsoft.copilot.eclipse.core.persistence.UserTurnData.MessageData;
 
 /**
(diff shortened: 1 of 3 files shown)


com.microsoft.copilot.eclipse.core/src/com/microsoft/copilot/eclipse/core/persistence/ConversationPersistenceManager.java
```diff
@@ -270,6 +272,41 @@ public CompletableFuture<Void> persistCachedConversation(String conversationId)
...
+    return CompletableFuture.runAsync(() -> {
+      lock.writeLock().lock();
+      try {
+        persistAndCacheConversation(conversationData);
+      } catch (IOException e) {
+        CopilotCore.LOGGER.error("Failed to persist cancelled tool calls for conversation: " + conversationId, e);
+      } finally {
+        lock.writeLock().unlock();
+      }
+    });
```
jdneo, line 307: I'm thinking that why not move the cache lookup, markRunningToolCallsCancelled(conversationData), and persistAndCacheConversation(conversationData) into the same CompletableFuture.runAsync() block while holding the write lock

Is there any concern to make them splited?

xinyi-gong: There wasn’t a strong reason. Updated.

jdneo (approved):

PR #247, opened 2026-05-21
feat - improve terminal command execution across windows and linux

xinyi-gong:
## Summary
This PR updates the Copilot terminal tool to use shell-aware command execution and marker-based completion across Windows and Linux.

## Behavior changes

- Switch Windows terminal execution from cmd.exe to powershell.exe
- Switch Linux terminal execution from /bin/sh to /bin/bash
- Remove POSIX sh shell integration support
- Add PowerShell and Bash shell integration scripts that emit Copilot OSC markers for:
                - prompt start
                - prompt end
                - command completion with exit code
- Use marker-based command completion instead of relying on prompt-character heuristics when shell integration is available
- Keep completion strict by requiring a command-finished marker followed by the next prompt-end marker
- Interrupt the previous foreground terminal command with Ctrl+C before starting a new foreground command
- Kill the active terminal when the user cancels the chat request
- Format multiline commands with bracketed paste markers
- Truncate terminal output returned to the model to the last 1000 lines
- Resolve terminal working directory from the current file’s project first, then referenced files
- Update terminal tool descriptions to match the actual shell


com.microsoft.copilot.eclipse.ui.terminal.tm/src/com/microsoft/copilot/eclipse/ui/terminal/tm/RunInTerminalTool.java
```diff
@@ -187,6 +215,69 @@ public StringBuilder getBackgroundCommandOutput(String executionId) {
...
+    }
+  }
+
+  private boolean hasRunningForegroundCommand() {
+    return StringUtils.isNotBlank(activeCommand) && resultFuture != null && !resultFuture.isDone();
+  }
+
+  private void sendInterrupt(ITerminalViewControl terminalViewControl) {
+    Display display = terminalViewControl.getControl().getDisplay();
+    display.syncExec(() -> {
```
jdneo: any reason sync is must?

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/tools/RunInTerminalToolAdapter.java
```diff
@@ -178,13 +197,63 @@ public CompletableFuture<LanguageModelToolResult[]> invoke(Map<String, Object> i
...
+    for (IResource resource : resources) {
+      String location = resolveProjectLocation(resource);
+      if (StringUtils.isNotBlank(location)) {
+        return location;
+      }
+    }
+    return "";
+  }
+
+  private static String resolveWorkingDirectory() {
```
jdneo, line 219: You can leverage ResourceUtils.deriveWorkspaceFoldersFrom()

xinyi-gong: Updated

com.microsoft.copilot.eclipse.ui.terminal.tm/src/com/microsoft/copilot/eclipse/ui/terminal/tm/RunInTerminalTool.java
xinyi-gong: Yes, `syncExec` is intentional here. When a new foreground command starts, i first send Ctrl+C to interrupt the currently running command, and then immediately paste the new command into the same reused terminal. If this were `asyncExec`, the interrupt could be queued behind the paste operation.

jdneo: ok

com.microsoft.copilot.eclipse.ui.terminal/src/com/microsoft/copilot/eclipse/ui/terminal/RunInTerminalTool.java
```diff
@@ -60,8 +68,9 @@ public class RunInTerminalTool implements IRunInTerminalTool {
   // Output and command state
   private StringBuilder sb;
   private CompletableFuture<String> resultFuture;
```
jdneo: `resultFuture` is read from the terminal output monitor thread but written from `executeCommand()` without being `volatile` or consistently protected by the same lock. The other command-state fields are volatile, so the monitor can observe the new command state while still seeing an old/null/completed future and skip completion detection for the current command. At minimum this should be `volatile`; preferably the future, active command, marker mode, and skip state should be captured in one immutable/synchronized command state so the monitor cannot observe a mixed state from two executions. The same issue exists in the TM terminal implementation copy.

com.microsoft.copilot.eclipse.terminal.api/src/com/microsoft/copilot/eclipse/terminal/api/TerminalCommandProcessor.java
```diff
@@ -0,0 +1,302 @@
...
+    }
+
+    // The command-finished marker is emitted before the next prompt. Wait for prompt end so the returned output keeps
+    // the prompt line, which gives the language model the terminal's current working directory.
+    MarkerRange promptEndMarkerRange = findMarker(output, PROMPT_END_MARKER_PATTERN, commandFinishMarkerRange.endIndex);
+    if (promptEndMarkerRange == null) {
+      return CompletionCheckResult.incomplete();
+    }
+
+    if (skipCompletion) {
```
jdneo: This skip is keyed only by a global boolean, so it discards the next complete marker region regardless of whether that region actually belongs to the interrupted command. One failure mode: Ctrl+C is sent for command A, `skipNextCompletionAfterInterrupt` is set, the buffer is cleared while A's stale marker is still pending, then command B completes; B's marker becomes the first complete region seen and is deleted as "skipped", leaving B's future waiting for later/unrelated output. This needs to be tied to a command generation/session marker or otherwise reset only when the interrupted command's marker is actually consumed.

com.microsoft.copilot.eclipse.ui.terminal/src/com/microsoft/copilot/eclipse/ui/terminal/RunInTerminalTool.java
xinyi-gong: Good point. I updated both terminal implementations to avoid observing mixed command state from the output monitor thread.

The per-command data is now captured in a single volatile `ForegroundCommand` snapshot. I intentionally kept `skipNextCompletionAfterInterrupt` outside of `ForegroundCommand` because it represents stale terminal-buffer cleanup after Ctrl+C rather than metadata for the newly running command. The skip behavior is handled separately in the completion processor.

com.microsoft.copilot.eclipse.terminal.api/src/com/microsoft/copilot/eclipse/terminal/api/TerminalCommandProcessor.java
xinyi-gong: Thanks, The completion processor now checks the completed region against the active command text before applying the skip. If the next completed region belongs to command B, it contains B's command text and will not be discarded. Only a stale completed region that does not contain the active command is skipped.

com.microsoft.copilot.eclipse.terminal.api/src/com/microsoft/copilot/eclipse/terminal/api/TerminalCommandProcessor.java
```diff
@@ -0,0 +1,314 @@
...
+
+  private static boolean shouldSkipCompletion(String completedRegion, String activeCommand, boolean skipCompletion) {
+    if (!skipCompletion) {
+      return false;
+    }
+    String normalizedCommand = normalizeLineEndings(activeCommand == null ? "" : activeCommand).trim();
+    if (normalizedCommand.isBlank()) {
+      return true;
+    }
+    return !normalizeLineEndings(completedRegion).contains(normalizedCommand);
```
jdneo: Will it happen that the interrupted command’s output happens to contain the next command text?

com.microsoft.copilot.eclipse.terminal.api/src/com/microsoft/copilot/eclipse/terminal/api/TerminalCommandProcessor.java
```diff
@@ -0,0 +1,314 @@
...
+
+  private static String cleanCommandOutput(String output, String activeCommand) {
+    String normalizedOutput = normalizeLineEndings(output);
+    normalizedOutput = removeShellIntegrationMarkers(normalizedOutput);
+    normalizedOutput = removeBracketedPasteMarkers(normalizedOutput);
+    String normalizedCommand = normalizeLineEndings(activeCommand == null ? "" : activeCommand).trim();
+    if (normalizedCommand.isBlank()) {
+      return normalizedOutput.trim();
+    }
+    int commandIndex = normalizedOutput.lastIndexOf(normalizedCommand);
```
jdneo: Using lastIndexOf(normalizedCommand) can discard valid stdout if the command text appears later in the command output. This should probably remove only the echoed command near the start of the completed region, e.g. by matching the first command echo line, rather than the last occurrence anywhere in the output.

com.microsoft.copilot.eclipse.ui.terminal.tm/src/com/microsoft/copilot/eclipse/ui/terminal/tm/RunInTerminalTool.java
```diff
@@ -265,80 +361,35 @@ private ITerminalServiceOutputStreamMonitorListener buildOutputStreamMonitorList
...
+    if (completionResult.state() == CompletionCheckState.INCOMPLETE) {
       return;
     }
-
-    String lastLine = terminalOutput.substring(lastNewLineIndex).trim();
-
-    // Check if last line is a prompt line
-    // Mac always has single '%' as last line, that's not what we want.
-    if (StringUtils.isBlank(lastLine) || lastLine.length() == 1) {
+    if (completionResult.state() == CompletionCheckState.SKIPPED) {
```
jdneo: Consider clearing `skipNextCompletionAfterInterrupt` on any non-incomplete completion result, not only when the result is `SKIPPED`.

If the interrupted command’s completion marker never arrives, or the new command completes first and returns `COMPLETED`, this flag can remain `true` and leak into a later command. That later command could then incorrectly skip a valid completion region.

For example:

```java
if (completionResult.state() != CompletionCheckState.INCOMPLETE) {
  skipNextCompletionAfterInterrupt = false;
}

if (completionResult.state() == CompletionCheckState.SKIPPED) {
  return;
}

File: com.microsoft.copilot.eclipse.terminal.api/src/com/microsoft/copilot/eclipse/terminal/api/TerminalCommandProcessor.java (hunk 5 above)
Comment (xinyi-gong) [reply com.microsoft.copilot.eclipse.terminal.api/src/com/microsoft/copilot/eclipse/terminal/api/TerminalCommandProcessor.java]: I simplified this flow and removed the command-text heuristic entirely.

We no longer send Ctrl+C and then reuse the same terminal for the next foreground command. If a foreground command is still running when a new one starts, the current Copilot terminal tab is closed and the command future is completed before the new terminal command is opened.

Because of that, the interrupted command's stale output/markers are not shared with the next command anymore. I also removed `skipNextCompletionAfterInterrupt`, `SKIPPED`, and the active-command matching logic from `TerminalCommandProcessor`.

File: com.microsoft.copilot.eclipse.terminal.api/src/com/microsoft/copilot/eclipse/terminal/api/TerminalCommandProcessor.java (hunk 6 above)
Comment (xinyi-gong) [reply com.microsoft.copilot.eclipse.terminal.api/src/com/microsoft/copilot/eclipse/terminal/api/TerminalCommandProcessor.java]: This logic was also removed

File: com.microsoft.copilot.eclipse.ui.terminal.tm/src/com/microsoft/copilot/eclipse/ui/terminal/tm/RunInTerminalTool.java (hunk 7 above)
Comment (xinyi-gong) [reply com.microsoft.copilot.eclipse.ui.terminal.tm/src/com/microsoft/copilot/eclipse/ui/terminal/tm/RunInTerminalTool.java]: Removed

File: com.microsoft.copilot.eclipse.terminal.api/src/com/microsoft/copilot/eclipse/terminal/api/TerminalCommandProcessor.java [hunk 8]
```diff
@@ -0,0 +1,269 @@
...
+    return false;
+  }
+
+  private static String removeTrailingLineEndings(String value) {
+    int endIndex = value.length();
+    while (endIndex > 0 && value.charAt(endIndex - 1) == '\n') {
+      endIndex--;
+    }
+    return value.substring(0, endIndex);
+  }
```
jdneo: Can we use `StringUtils.stripEnd` here?

com.microsoft.copilot.eclipse.terminal.api/src/com/microsoft/copilot/eclipse/terminal/api/TerminalCommandProcessor.java
```diff
@@ -0,0 +1,269 @@
...
+    normalizedOutput = removeShellIntegrationMarkers(normalizedOutput);
+    normalizedOutput = removeBracketedPasteMarkers(normalizedOutput);
+    return normalizedOutput.trim();
+  }
+
+  private static String removeShellIntegrationMarkers(String output) {
+    return output.replaceAll(ShellIntegrationScripts.OSC_MARKER_PATTERN, "")
+        .replace(ShellIntegrationScripts.PROMPT_START_MARKER, "")
+        .replace(ShellIntegrationScripts.PROMPT_END_MARKER, "")
+        .replace(ShellIntegrationScripts.COMMAND_FINISH_MARKER, "");
```
jdneo: Q: Are those already covered by .replaceAll(ShellIntegrationScripts.OSC_MARKER_PATTERN, "")?

com.microsoft.copilot.eclipse.ui.terminal/src/com/microsoft/copilot/eclipse/ui/terminal/RunInTerminalTool.java
```diff
@@ -71,64 +76,71 @@ public RunInTerminalTool() {
...
       return CompletableFuture.completedFuture("The command is null or empty.");
     }
 
-    resultFuture = new CompletableFuture<>();
-
-    // Linux uses shell-integration marker lines; others rely on prompt detection
-    useMarker = Platform.getOS().equals(Platform.OS_LINUX);
+    if (!isBackground) {
+      closeRunningForegroundTerminal(COMMAND_INTERRUPTED_MESSAGE);
+    }
```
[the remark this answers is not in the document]
xinyi-gong: This is intentional

com.microsoft.copilot.eclipse.ui.terminal/src/com/microsoft/copilot/eclipse/ui/terminal/RunInTerminalTool.java
```diff
@@ -71,64 +76,71 @@ public RunInTerminalTool() {
...
-      // For marker-based detection, clear the buffer
-      sb.setLength(0);
     }
 
     String executionId = UUID.randomUUID().toString();
-    final String finalCommand = command + System.lineSeparator();
+    final String finalCommand = TerminalCommandProcessor.formatForExecution(command, useBracketedPaste());
 
     synchronized (lock) {
       if (!isBackground && this.persistentTerminalViewControl != null) {
```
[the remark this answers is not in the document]
xinyi-gong: This is intentional for the persistent terminal behavior.

com.microsoft.copilot.eclipse.terminal.api/scripts/copilot-powershell-integration.ps1
```diff
@@ -0,0 +1,52 @@
...
+    function global:prompt {
+        $lastSuccess = $?
+        $lastExitCode = $LASTEXITCODE
+        $esc = [char]27
+        $bel = [char]7
+        $lastHistoryEntry = Get-History -Count 1
+        $result = ""
+
+        if ($global:__copilot_last_history_id -ne -1) {
+            if ($lastHistoryEntry.Id -eq $global:__copilot_last_history_id) {
```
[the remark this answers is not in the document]
xinyi-gong: Updated

com.microsoft.copilot.eclipse.terminal.api/src/com/microsoft/copilot/eclipse/terminal/api/TerminalCommandProcessor.java
xinyi-gong: Updated. Added the StringUtils dependency in MANIFEST

com.microsoft.copilot.eclipse.terminal.api/src/com/microsoft/copilot/eclipse/terminal/api/TerminalCommandProcessor.java
```diff
@@ -0,0 +1,269 @@
...
+        .append(" lines.]\n");
+    for (int lineIndex = firstLineToKeep; lineIndex < lineCount; lineIndex++) {
+      if (lineIndex > firstLineToKeep) {
+        truncatedOutput.append('\n');
+      }
+      truncatedOutput.append(lines[lineIndex]);
+    }
+    if (endsWithNewline) {
+      truncatedOutput.append('\n');
+    }
```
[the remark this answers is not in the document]
xinyi-gong: Updated. `truncateOutput` now scans from the end until it finds the tail boundary and returns a substring, so it avoids allocating an array/string per output line.

com.microsoft.copilot.eclipse.terminal.api/src/com/microsoft/copilot/eclipse/terminal/api/TerminalCommandProcessor.java
xinyi-gong: Yes, those are covered by `OSC_MARKER_PATTERN`. I removed the redundant explicit marker replacements.

com.microsoft.copilot.eclipse.terminal.api/src/com/microsoft/copilot/eclipse/terminal/api/TerminalCommandProcessor.java
```diff
@@ -0,0 +1,269 @@
...
+    while (matcher.find()) {
+      output.delete(matcher.start(), matcher.end());
+      matcher.reset(output);
+    }
+  }
+
+  private static Pattern buildMarkerPattern(String markerKind, boolean includeExitCode) {
+    String exitCodePattern = includeExitCode ? "(?:;[-]?\\d+)?" : "";
+    return Pattern.compile("(?:\u001B)?\\]" + ShellIntegrationScripts.OSC_NAMESPACE + ";" + markerKind
+        + exitCodePattern + "(?:\u0007|\u001B\\\\)?");
```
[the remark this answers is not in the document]
xinyi-gong: This partial-marker case should be very unlikely in practice, since the marker is emitted by our shell integration and completion still waits for the following prompt-end marker.

I kept the terminator optional because Eclipse terminal output can strip OSC framing bytes, and we need to support bare markers like `]7775;C;0]7775;A... ]7775;B`. Requiring BEL/ESC\ would break that stripped-marker case.

jdneo (approved):
LGTM

jdneo:
@iloveeclipse There are some improvement about the run in terminal tool in latest nightly. Please let me know if you meet any unexpected behavior

PR #250, opened 2026-05-21
feat: Implement auto compression feature for chat context

ethanyhou:
**Demo:**

https://github.com/user-attachments/assets/efef11cf-f3f1-4b69-8b7e-bd27d282d10c

The status context aligns with VSCode:
image


com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/BaseTurnWidget.java
```diff
@@ -234,6 +234,32 @@ public void appendMessage(String message) {
...
+    // Defer rendering when the partial could be a code-block fence:
+    //   - "```..." → confirmed fence prefix, wait for the newline before deciding.
+    //   - "`" or "``" with no other content → could grow into a fence on the next chunk.
+    if (trimmed.startsWith(CODE_BLOCK_ANNOTATION)
+        || (trimmed.length() <= 2 && trimmed.chars().allMatch(c -> c == '`'))) {
+      return;
+    }
+    SwtUtils.invokeOnDisplayThread(() -> {
+      appendTextToTextViewer(mdContentBuilder.toString() + partial);
+    }, this);
```
ethanyhou: This part is to fix a corner case: when the compacting conversation happens the last reply message cannot be rendered. Render partial buffer can force the pending message to render on the chat view.

jdneo:
Any consideration to expose a preference to disable auto compact?

I prefer to always enable it, enable it simply give users more value and advantages.

iloveeclipse:
> Any consideration to expose a preference to disable auto compact?
> 
> I prefer to always enable it, enable it simply give users more value and advantages.

We have errors reported by customers due the context "overflow" with Claude models, having this feature enabled by default would probably help to avoid the errors. What would be the point in *not* enabling it by default? Just the fear that the feature doesn't work properly or are there some use cases where it would harm if enabled?

If there is no use case for disabling the auto compression, I would hide the preference or just don't provide one at all.

ethanyhou:
> Any consideration to expose a preference to disable auto compact?
> 
> I prefer to always enable it, enable it simply give users more value and advantages.

@jdneo Removed as we aligned.

ethanyhou:
Hi @iloveeclipse,

Thanks for the input. Agreed and it was removed. Compression will always be enabled.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/BaseTurnWidget.java
```diff
@@ -234,6 +234,32 @@ public void appendMessage(String message) {
...
     }
+    // Eagerly render any trailing partial line (text without a closing newline) so the user
+    // sees streamed content immediately instead of waiting for the next chunk or a turn-end
+    // flush. The markup viewer is fully overwritten on each setMarkup call, so the next
+    // chunk (which either completes the line or adds more partial text) will overwrite this
+    // tentative render with the committed content. Skip when inside a code block, because
+    // partial text must be routed to the source viewer, and when the partial could be the
+    // start of a code-block fence, because fence detection requires a complete line.
+    renderPartialBuffer();
+  }
```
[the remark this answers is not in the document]
ethanyhou: Added.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/BaseTurnWidget.java
```diff
@@ -234,6 +234,32 @@ public void appendMessage(String message) {
...
       processMessageLine(line);
     }
+    // Eagerly render any trailing partial line (text without a closing newline) so the user
+    // sees streamed content immediately instead of waiting for the next chunk or a turn-end
+    // flush. The markup viewer is fully overwritten on each setMarkup call, so the next
+    // chunk (which either completes the line or adds more partial text) will overwrite this
+    // tentative render with the committed content. Skip when inside a code block, because
+    // partial text must be routed to the source viewer, and when the partial could be the
+    // start of a code-block fence, because fence detection requires a complete line.
+    renderPartialBuffer();
```
jdneo: Curious how do you come up with this? Are there any case breaking you without this fix?

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatView.java
```diff
@@ -394,6 +398,53 @@ public void done(IJobChangeEvent event) {
     };
     this.eventBroker.subscribe(CopilotEventConstants.TOPIC_QUOTA_WARNING, this.quotaWarningHandler);
 
+    this.compressionStartedHandler = event -> {
+      Object data = event.getProperty(IEventBroker.DATA);
+      if (!(data instanceof CompressionStartedParams params)) {
+        return;
+      }
+      if (!StringUtils.equals(params.conversationId(), this.conversationId)) {
```
jdneo: we have conversation id and subagent conversation id. What will happen when compact happens in a subagent? Which id is used?

ethanyhou: This is on purpose. The subagent's conversationId is logically a different conversation. When the LSP emits `compressionStarted/Completed` for it, the payload describes the subagent's state, not the parent's to guard compression is doing the real work, not accidentally swallowing an event.

Also in Jetbrains:
https://github.com/microsoft/copilot-intellij/blob/cee23502c288e338a4c596b3cce1e027ed800e56/core/src/main/kotlin/com/github/copilot/agent/session/component/CopilotAgentSessionComponent.kt#L1217-L1235

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/BaseTurnWidget.java
ethanyhou: This is to fix the error I mentioned above: https://github.com/microsoft/copilot-for-eclipse/pull/250#discussion_r3279885103

I didn't found regression from the current implementation.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatView.java
jdneo: I mean, if CLS returns the subagent's conversation id, we will miss that here.

@duzitong Could you clarify the CLS behavior?

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/BaseTurnWidget.java
```diff
@@ -240,6 +240,32 @@ public void appendMessage(String message) {
...
```
jdneo: can we use notify end and get rid of this?

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatView.java
ethanyhou: @jdneo Added as we aligned.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/BaseTurnWidget.java
ethanyhou: Removed directly.

The buffer will be flushed by the code that added in `ChatView.java` around line 1437:
```
if (this.chatContentViewer != null && !this.chatContentViewer.isDisposed()) {
      CopilotTurnWidget turn = this.chatContentViewer.getLatestCopilotTurn();
      if (turn != null) {
        // Flush any buffered reply text (e.g. a final line that arrived without a trailing
        // newline) so it is rendered before we hide the compacting banner. Any subsequent
        // end progress event from the language server is dropped because the action bar has
        // already been reset to the send state, so this is the only chance to render it.
        turn.notifyTurnEnd();
        turn.hideCompactingStatus();
        // Refresh the scroll layout so the ScrolledComposite picks up the new content size
        // produced by notifyTurnEnd/hideCompactingStatus; otherwise the just-flushed reply
        // can be clipped or invisible until the next layout pass.
        this.chatContentViewer.refreshScrollerLayout();
      }
    }
```

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatView.java
```diff
@@ -396,6 +400,53 @@ public void done(IJobChangeEvent event) {
...
+      }
+      if (!isCompressionForActiveConversation(params.conversationId())) {
+        return;
+      }
+      SwtUtils.invokeOnDisplayThreadAsync(() -> {
+        if (this.chatContentViewer == null || this.chatContentViewer.isDisposed()) {
+          return;
+        }
+        CopilotTurnWidget turn = this.chatContentViewer.getLatestCopilotTurn();
+        if (turn != null) {
```
jdneo: Please double check if we need to call notifyTurnEnd here?

ethanyhou: No.
Checked the CLS side, the Eclipse side buffer is empty before CLS calling the `notifyCompressionStarted`.

1. In turnProcessor.ts and agenticTurnProcessor.ts `checkAndCompress` runs before the LLM request is dispatched. At this point the `begin` progress has fired but no `report` reply text has streamed yet, so the Copilot turn's `messageBuffer` is empty.
2. In toolCallingLoop.ts `checkAndCompress` runs after a tool round completes. On the Eclipse side, that round's preceding `appendToolCallStatus` already invoked `reset()` in `BaseTurnWidget.java` which flushes the message buffer before showing the tool status.

jdneo: Not true, one easy way to verify it is:

1. Change the CLS side logic, to make it kick off compression when context window consumption is larger than 20%. (Or other number to make it be triggered easier)
2. Add notifyTurnEnd() when compression start
   
Screenshot 2026-05-27 at 12 46 00
3. Chat with agent to trigger compresision, you will see when breakpoint hits, the buffer contains chars:
image

ethanyhou: Verified and added. Thanks

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatView.java
```diff
@@ -396,6 +400,57 @@ public void done(IJobChangeEvent event) {
...
+          turn.hideCompactingStatus();
+          this.chatContentViewer.refreshScrollerLayout();
+        }
+        if (params.contextInfo() != null) {
+          this.chatServiceManager.getContextWindowService().updateContextSize(params.contextInfo());
+        }
+      }, parent);
```
jdneo, line 437: Can we move this section into the chatContentViewer? Then no need for chatContentViewer to expose the method `getLatestCopilotTurn()`

Same ask for compressionStartedHandler

ethanyhou: Updated.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatContentViewer.java
```diff
@@ -380,11 +380,41 @@ private void processTodoListFromToolCall(ChatServiceManager chatServiceManager,
   }
 
   /**
-   * Returns the most-recent Copilot turn widget, or {@code null} if none exists or it has been disposed.
+   * Shows the compacting status on the latest Copilot turn after flushing any buffered reply text.
+   */
+  public void showCompactingStatusOnLatestCopilotTurn() {
+    CopilotTurnWidget turn = getLatestCopilotTurn();
```
jdneo: latestCopilotTurn is member, which can directly access, I don't see the benefit introduce `getLatestCopilotTurn()`, and this is why I propose to move logics into ChatContentViewer in https://github.com/microsoft/copilot-for-eclipse/pull/250#discussion_r3308999373

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatContentViewer.java
```diff
@@ -380,11 +380,41 @@ private void processTodoListFromToolCall(ChatServiceManager chatServiceManager,
...
+    if (turn == null) {
```
jdneo: If using latestCopilotTurn directly, check dispose status here.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatContentViewer.java
```diff
@@ -380,11 +380,41 @@ private void processTodoListFromToolCall(ChatServiceManager chatServiceManager,
...
+      return;
+    }
+    // Flush any buffered reply text from the previous round so it is rendered
+    // above the compacting spinner; otherwise it would be concatenated with
+    // the next round's reply and produce a single garbled line.
+    turn.notifyTurnEnd();
```
jdneo: I'm considering to rename notifyTurnEnd(). It's a little bit confused now since compression can happen in the middle of the turn. How about `flushMessageBuffer()`

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatContentViewer.java
```diff
@@ -380,11 +380,41 @@ private void processTodoListFromToolCall(ChatServiceManager chatServiceManager,
...
+  public void hideCompactingStatusOnLatestCopilotTurn(boolean flushBufferedReply) {
+    CopilotTurnWidget turn = getLatestCopilotTurn();
+    if (turn == null) {
+      return;
+    }
+    if (flushBufferedReply) {
+      // Flush any buffered reply text before hiding the compacting banner; callers that
+      // cancel a turn may not receive a later end progress event to do this for them.
+      turn.notifyTurnEnd();
+    }
```
jdneo: Let's make it simpler, hideCompactingStatusOnLatestCopilotTurn will always flush the buffer (which should have very tiny load because the buffer should be empty at this stage. If the buffer is not empty at this stage for some unknown reason, this can be a guard)

That way, `boolean flushBufferedReply` can be removed

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatContentViewer.java
ethanyhou: Addressed

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatContentViewer.java
ethanyhou: Addressed

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatContentViewer.java
ethanyhou: Addressed

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatContentViewer.java
ethanyhou: Addressed

jdneo (approved):

PR #251, opened 2026-05-22
docs: fix typo in README upgrade recommendation

evanclan:
## What

Fix a small typo in `README.md` under the **Usage-based billing support** section.

The current wording contains a duplicated `to`:

> we strongly recommend upgrading **to** the plugin **to** **0.18.0 or later** as soon as possible.

This change drops the redundant first `to` so the sentence reads naturally:

> we strongly recommend upgrading the plugin to **0.18.0 or later** as soon as possible.

## Why

The wording was introduced in #227 when the usage-based billing section was added. The duplicated `to` makes the sentence grammatically incorrect and is the first thing users see in the README, so it is worth tightening up.

## Scope

- Single-line, docs-only change. No source, build, test, or behavior is affected.
- No CLA-relevant code change; pure prose fix.

## Verification

- `git diff` shows a one-character (well, three-character) change confined to `README.md`.
- `README.md` still renders correctly as Markdown; no link, list, heading, or code block was touched.

Made with [Cursor](https://cursor.com)


jdneo (approved):

PR #252, opened 2026-05-23
fix: refresh scroller layout when tool confirmation dialog is cancelled

rsd-darshan:
## Summary

- When the user cancels a subagent tool confirmation dialog, the subagent panel stays expanded at its pre-cancel height
- Root cause: `cancelConfirmation()` calls `parent.layout()` (relays the turn widget's children) but never updates the `ScrolledComposite`'s minimum size via `refreshScrollerLayout()`
- On the Continue path this goes unnoticed because subsequent subagent messages trigger a layout refresh; on the Cancel path no further messages arrive, so the panel is stuck

## Fix

In `InvokeToolConfirmationDialog.cancelConfirmation()`, after `parent.layout()`, walk up the ancestor chain to find `ChatContentViewer` and call `refreshScrollerLayout()`. This recomputes `setMinHeight()` so the scroll area collapses correctly after cancellation.

## Test plan

- [ ] Invoke a subagent (e.g. "scan for CVEs")
- [ ] Wait for the tool confirmation dialog to appear
- [ ] Click **Cancel**
- [ ] Verify the subagent panel resizes/collapses correctly and does not stay stuck at the expanded height
- [ ] Click **Continue** on a separate invocation and verify that path is unaffected


rsd-darshan:
Closing to reopen from a clean branch based on main — this branch accidentally included unrelated commits.

PR #254, opened 2026-05-23
fix: refresh scroller layout when tool confirmation dialog is cancelled

### LINKED ISSUE #169: [Bug] Subagent panel size is not updated if cancel in the middle
1. Try to invoke a subagent, i.e. ask :"scan for CVEs"
2. When the subagent wants to run a command, wait for the confirmation dialog appear
3. Click cancel

An empty area can be observed at the end of subagent area:

Image

rsd-darshan:
## Summary

- When the user cancels a subagent tool confirmation dialog, the subagent panel stays expanded at its pre-cancel height
- Root cause: `cancelConfirmation()` calls `parent.layout()` (relays the turn widget's children) but never updates the `ScrolledComposite`'s minimum size via `refreshScrollerLayout()`
- On the Continue path this goes unnoticed because subsequent subagent messages trigger a layout refresh; on the Cancel path no further messages arrive, so the panel is stuck

## Fix

Pass a `Runnable` callback from `BaseTurnWidget` into `InvokeToolConfirmationDialog` at construction. `BaseTurnWidget` walks up the ancestor chain once to find `ChatContentViewer` and captures `refreshScrollerLayout()` as the callback. `cancelConfirmation()` calls it after disposal, which recomputes `setMinHeight()` in a single layout pass and collapses the panel correctly. This also removes the redundant `parent.layout()` call that preceded the deep `layout(true,true)` inside `refreshScrollerLayout()`.

## Test plan

- [x] Invoke a subagent (e.g. "run ls /Users/darshanpoudel/Desktop in the terminal")
- [x] Wait for the tool confirmation dialog to appear
- [x] Click **Cancel**
- [x] Verify the subagent panel resizes/collapses correctly and does not stay stuck at the expanded height
- [x] Click **Continue** on a separate invocation and verify that path is unaffected

Fixes #169

### DIFF
--- a/com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/BaseTurnWidget.java
+++ b/com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/BaseTurnWidget.java
@@ -649,17 +649,17 @@ public abstract class BaseTurnWidget extends Composite {
       ConfirmationContent content, Object input) {
     reset();
 
-    Runnable refreshLayout = null;
-    Composite ancestor = this.getParent();
-    while (ancestor != null && !ancestor.isDisposed()) {
-      if (ancestor instanceof ChatContentViewer) {
-        final ChatContentViewer viewer = (ChatContentViewer) ancestor;
-        refreshLayout = viewer::refreshScrollerLayout;
-        break;
+    this.confirmDialog = new InvokeToolConfirmationDialog(this, content, input);
+    this.confirmDialog.addDisposeListener(e -> {
+      Composite ancestor = this.getParent();
+      while (ancestor != null && !ancestor.isDisposed()) {
+        if (ancestor instanceof ChatContentViewer) {
+          ((ChatContentViewer) ancestor).requestRefreshScrollerLayout();
+          break;
+        }
+        ancestor = ancestor.getParent();
       }
-      ancestor = ancestor.getParent();
-    }
-    this.confirmDialog = new InvokeToolConfirmationDialog(this, content, input, refreshLayout);
+    });
     CompletableFuture<LanguageModelToolConfirmationResult> toolConfirmationFuture = this.confirmDialog
         .getConfirmationFuture();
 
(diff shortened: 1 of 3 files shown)


jdneo:
Thanks for tracking down the root cause. I agree that the missing piece is recomputing the `ScrolledComposite` min height after the cancel path, since no later message arrives to trigger `refreshScrollerLayout()`.

I would prefer not to inject a layout callback into `InvokeToolConfirmationDialog`, though. That makes a leaf confirmation widget aware of the chat container's layout policy. Could we split the responsibilities instead?

- Keep `InvokeToolConfirmationDialog` self-contained: both accept and cancel paths can use a private helper like `disposeAndRequestParentLayout()`, which disposes the dialog and calls `parent.requestLayout()` on its immediate parent.
- Let `ChatContentViewer` own a coalesced async API such as `requestRefreshScrollerLayout()`, which schedules a single `refreshScrollerLayout()` after pending SWT mutations settle.
- In `BaseTurnWidget.requestToolExecutionConfirmation(...)`, where the dialog is created, add a dispose listener on the dialog that walks to the nearest `ChatContentViewer` and calls `requestRefreshScrollerLayout()`.

That should fix the stuck subagent panel for cancel, accept, and programmatic cancellation uniformly, while keeping scroller-specific behavior out of `InvokeToolConfirmationDialog`.

rsd-darshan:
@jdneo Thanks for the detailed feedback — that's a much cleaner split of responsibilities.

I've updated the implementation:

- `InvokeToolConfirmationDialog` is now self-contained: both accept and cancel paths use a private `disposeAndRequestParentLayout()` helper that disposes the dialog and calls `parent.requestLayout()` on its immediate parent. No layout policy leaks out of the dialog.
- `ChatContentViewer` gains a `requestRefreshScrollerLayout()` method that schedules a single async `refreshScrollerLayout()` call, coalescing multiple dispose/layout events that arrive in the same event-loop tick.
- `BaseTurnWidget.requestToolExecutionConfirmation()` attaches a dispose listener on the dialog that walks up to the nearest `ChatContentViewer` and calls `requestRefreshScrollerLayout()`. This covers cancel, accept, and programmatic cancellation uniformly.

jdneo (approved):

PR #256, opened 2026-05-25
fix -  nested scrolling and full-row table selection for auto-approve preferences

xinyi-gong:
## Summary

- Forward vertical mouse wheel events from auto-approve inner viewers to the parent preference scroller only after the inner viewer reaches its scroll boundary
- Update auto-approve tables so the last column fills remaining table width, matching the custom instructions table behavior


jdneo (approved):
LGTM

PR #266, opened 2026-05-28
fix: Request layout update after setting markup in AgentStatusLabel

ethanyhou:
1. `setMinSize` already called layout, force layout in `refreshScrollerLayout` after `setMinSize` is expensive and redundant, should be removed.
2. `AgentStatusLabel` should requestLayout after text changes.

### DIFF
--- a/com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/AgentStatusLabel.java
+++ b/com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/AgentStatusLabel.java
@@ -142,6 +142,7 @@ public class AgentStatusLabel extends Composite {
       AccessibilityUtils.addFocusBorderToComposite(styledText);
     }
     textLabel.setMarkup(text);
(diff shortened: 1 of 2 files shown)


jdneo:
@ethanyhou Do you have any data point about how perf boost can get with this change?

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/AgentStatusLabel.java
```diff
@@ -142,6 +142,7 @@ public void setText(String text) {
       AccessibilityUtils.addFocusBorderToComposite(styledText);
     }
     textLabel.setMarkup(text);
```
jdneo, line 144: @ethanyhou Could you help me remember why we are using markup viewer for the status label?

jdneo (approved):

jdneo:
The `layout(true, true)` is eliminated by https://github.com/microsoft/copilot-for-eclipse/pull/321.

Closing this one.

Issue #238, opened 2026-05-19
[Feat] Have option to use tabs instead of spaces (forced override)


jomillerOpen:
**Is your feature request related to a problem? Please describe.**
A problem is that the current core.format options do not support all IDE build environments. These 4 Java files will check for the preferences based on the type of project being worked on (defaulting to use spaces), but if you aren't using one of those project types, spaces will be used. These formatter options are used by the auto completions for code.

**Describe the solution you'd like**
Have a new checkbox on the completions preferences page to allow for an override to use tabs instead of spaces for indentation. When checked, this will override the format options that were determined/found and use tabs for auto completion code.

**Describe alternatives you've considered**
Have currently replaced the format options to use tabs instead of spaces by default. This means though manually updating the JAR file for each update. Alternatively it means trying to find/replace after finishing in a file 4 spaces for a tab character.

**Additional context**
The 4 format java files under com.microsoft.copilot.eclipse.core.format that use "space" instead of "tab" as default. Not sure if the code in these if fully functional (as in Java and workspace preferences, the tab policy is tabs only and "use spaces instead of tabs" is unchecked). Do to the way these formatter options are read, it ends up using spaces instead of tabs for new code blocks for auto-completions. For those that use tabs for indentation, not having an override option means a lot of fixing of code afterwards or mismatched indentation in files.

jdneo:
Hi @jomillerOpen

Could you provide repro steps for this issue?

jomillerOpen:
Hi @jdneo

From my quick debugging, the entry point to determine the tab setting is in FormatOptionProvider.java's useSpace() function. This then calls getLanguageFormat(), which goes off to try to determine the language being worked with and get it's preference value. The only language formatters it supports is Java or C/C++. Anything else ends up getting down to the lines: 
if (languageFormatReaderForProject == null) {
        return new FormattingOptions(DEFAULT_TAB_SIZE, DEFAULT_USE_SPACE);
      }
Where it will default to using spaces. I know the comment mentions using workspace preferences, but there isn't any code it seems that should read the generic editor tab options (Preferences>General>Editors>Text Editors: Insert spaces for tabs & Displayed tab width).

If we wanted to use that setting as the backup (instead of having a new checkbox/preference to control it), we should really be trying to get those settings if the languageFormatReaderForProject is null and then gettting the "spacesForTabs" and "tabSize" preference values.

[...]

PR #267, opened 2026-05-28
[Feat] Read the default text editor preferences from platform instead of defaulting to spaces.

jomillerOpen:
Read the default text editor preferences from Eclipse instead of defaulting to spaces by default. To avoid having additional plug-in dependencies, we use the existing preferences plug-in to check the multiple workspace instances for the settings. The 2 settings are the "spacesForTabs" and "tabWidth". If those settings are found, we will use those for the formatting options. If not, then it will continue to fall back to using defaults of using spaces. These settings are in effect when a project is neither a Java or C project type.

### DIFF
--- a/com.microsoft.copilot.eclipse.core.test/src/com/microsoft/copilot/eclipse/core/completion/FormatOptionProviderTests.java
+++ b/com.microsoft.copilot.eclipse.core.test/src/com/microsoft/copilot/eclipse/core/completion/FormatOptionProviderTests.java
@@ -10,6 +10,8 @@ import static org.mockito.Mockito.when;
 
 import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IProject;
+import org.eclipse.core.runtime.Platform;
+import org.eclipse.core.runtime.preferences.IPreferencesService;
 import org.eclipse.lsp4j.FormattingOptions;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
@@ -25,7 +27,11 @@ class FormatOptionProviderTests {
   private IFile mockFile;
   private IProject mockProject;
 
+  private static final String EDITOR_PREF_NODE = "org.eclipse.ui.editors";
+  private static final String TAB_WIDTH_KEY = "tabWidth";
+  private static final String SPACES_FOR_TABS_KEY = "spacesForTabs";
   private static final int PREFERENCE_DEFAULT_TAB_SIZE = 4;
+  private static final boolean PREFERENCE_DEFAULT_USE_SPACE = true;
 
   @BeforeEach
   void setUp() {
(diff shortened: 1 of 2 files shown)


jomillerOpen:
@microsoft-github-policy-service agree company="OpenText"

jomillerOpen:
Implements/satisfies #238

com.microsoft.copilot.eclipse.core/src/com/microsoft/copilot/eclipse/core/format/FormatOptionProvider.java
```diff
@@ -110,14 +116,49 @@ private FormattingOptions getLanguageFormat(IFile file) {
...
   }
 
+  /**
+   * Attempts to read the Eclipse default text editor preferences for tab size and spaces-for-tabs.
+   * Falls back to hardcoded defaults if the preferences cannot be read.
+   */
+  private FormattingOptions getEclipseTextEditorFormattingOptions() {
+    try {
+      Preferences instancePrefs = InstanceScope.INSTANCE.getNode(EDITOR_PREF_NODE);
+      Preferences defaultPrefs = DefaultScope.INSTANCE.getNode(EDITOR_PREF_NODE);
```
jdneo: Can we use Platform.getPreferencesService() here? which help us handles different scopes (Instance, default, configuration).

jomillerOpen:
I didn't know about that until now! But yes, I tried using it and it seems to work find as well. I'll switch it to that.

jdneo:
@jomillerOpen Looks good! Please fix the test failure, thank you.

> FormatOptionProviderTests.testGetCopilotDefaultTabCharAndSizeForUnknownLanguage:59 expected: <true> but was: <false>

jdneo:
CI still failed, pls refer to https://github.com/microsoft/copilot-for-eclipse/pull/273

jomillerOpen:
There we go. Passed locally. Had to return the Eclipse options for the 2 other scenarios where a file isn't part of a handled project.

com.microsoft.copilot.eclipse.core.test/src/com/microsoft/copilot/eclipse/core/completion/FormatOptionProviderTests.java
```diff
@@ -56,16 +62,38 @@ void testGetEclipseDefaultJavaTabCharAndSize() {
...
+          EDITOR_PREF_NODE,
+          TAB_WIDTH_KEY,
+          PREFERENCE_DEFAULT_TAB_SIZE,
+          null
+      );
+      return new FormattingOptions(tabSize, useSpaces);
+    } catch (Exception e) {
+      return new FormattingOptions(PREFERENCE_DEFAULT_TAB_SIZE, PREFERENCE_DEFAULT_USE_SPACE);
+    }
+  }
```
jdneo: The new `getEclipseTextEditorFormattingOptions()` helper duplicates the same preference lookup logic that the production code is expected to use. That makes the tests less effective, because if the implementation uses the wrong preference node/key/default, the test may still pass as long as this helper has the same mistake.

Could we instead set explicit editor preferences in the test and assert fixed expected values? For example:

```java
import org.eclipse.core.runtime.preferences.IEclipsePreferences;
import org.eclipse.core.runtime.preferences.InstanceScope;

private static final String EDITOR_PREF_NODE = "org.eclipse.ui.editors";
private static final String TAB_WIDTH_KEY = "tabWidth";
private static final String SPACES_FOR_TABS_KEY = "spacesForTabs";

@Test
void testGetCopilotDefaultTabCharAndSizeForUnknownLanguage() {
  when(mockFile.getFileExtension()).thenReturn("js");

  setEditorFormattingPreferences(false, 2);

  assertFalse(formatOptionProvider.useSpace(mockFile));
  assertEquals(2, formatOptionProvider.getTabSize(mockFile));
}

@Test
void testGetCopilotDefaultTabCharAndSizeForNoExtensionFile() {
  when(mockFile.getFileExtension()).thenReturn(null);

  setEditorFormattingPreferences(false, 2);

  assertFalse(formatOptionProvider.useSpace(mockFile));
  assertEquals(2, formatOptionProvider.getTabSize(mockFile));
}

private void setEditorFormattingPreferences(boolean useSpaces, int tabSize) {
  IEclipsePreferences prefs = InstanceScope.INSTANCE.getNode(EDITOR_PREF_NODE);
  prefs.putBoolean(SPACES_FOR_TABS_KEY, useSpaces);
  prefs.putInt(TAB_WIDTH_KEY, tabSize);
}
```

Optionally, the two test cases could be combined with a parameterized test:

```java
@ParameterizedTest
@NullSource
@ValueSource(strings = { "js" })
void testUsesEclipseTextEditorFormattingOptionsForUnknownOrNoExtension(String extension) {
  when(mockFile.getFileExtension()).thenReturn(extension);

  setEditorFormattingPreferences(false, 2);

  assertFalse(formatOptionProvider.useSpace(mockFile));
  assertEquals(2, formatOptionProvider.getTabSize(mockFile));
}
```

This keeps the test focused on observable behavior instead of mirroring the implementation’s preference-reading logic.

jomillerOpen: Updated with the parameterized test option and set to non-default options. Locally passed/worked.

jdneo (approved):
LGTM, thank you for your contribution!

Issue #143, opened 2026-05-07
Unify all .md resources loading approach


jdneo:
See: https://github.com/microsoft/copilot-for-eclipse/pull/136#discussion_r3199616948

Currently we have

<img width="869" height="324" alt="Image" src="https://github.com/user-attachments/assets/8d83d664-d077-4d81-80fa-7fceb4632a2b" />

In the preference.

An idea is make it affect to other md files as well. For example the skills.

And we can extract it out of the custom instructions block, because it ultimately affects how the workspace folders are passed to CLS. And CLS searches those md files based on the passed folder.

travkin79:
I'm trying to collect all the details that we need to consider and some open questions to answer.

**Files to consider**
* `copilot-instructions.md`
* `*.instructions.md`
* `*.prompt.md`
* `SKILL.md`
* `*.agent.md`
* `AGENTS.md`
* `CLAUDE.md`, `GEMINI.md`

**Copilot scopes according to documentation**
* **repository-wide** (files in git repository root folder or in `.github/`, `.claude/`, or `.agents/` folders)
* project- / **path-specific** (files in a project's directory or its sub-folders)
* **user-specific** / local (files in `~/.copilot/`, `~/.claude/`, or `~/.agents/`)

**Scopes in copilot-for-eclipse**
* instructions from **all** Eclipse projects in Eclipse workspace
* instructions only from Eclipse projects derived **from referenced files** in context (selected in Chat View)

**Current state**
* Scope for loading `copilot-instructions.md` can now be set to **all** (from all projects) or **from referenced files** when using the chat view (see #136)
* Scope for loading `*.prompt.md` and `SKILL.md` is always **all** (from all projects)
* Scope for loading `*.agent.md` is always **all** (from all projects)
* The scope settings also apply to the search (by limiting the set of workspace folders passed to CLS)

**Questions**
* Which of the listed files need scoping and which should always be loaded? (E.g. always load all user-specific instructions/skills?)
* Which scope settings do we need in copilot-for-eclipse?
* Do we need a separate scope setting for the search feature (e.g. loading instructions from a limited scope, but search in all projects in the Eclipse workspace)? Is that even possible or would that require changes on LS-side?
* Where should we place the new / adapted preferences (which preferences page)?
* Does the CLS load user-specific and/or repository-wide instructions or should loading them be a client-side feature?

**Considerations**
As a user, I'd expect copilot to
[...]

PR #269, opened 2026-05-29
Fix prompt and skills discovery at workspace root level

### LINKED ISSUE #245: [Bug] Workspace prompts and skills not recognized by plugin (ADT/Eclipse)
**Environment**
- OS: Windows 11
- Eclipse Version: 2026-03 + ADT
- Plugin Version: 18

**Describe the bug**
The plugin does not load the prompts and skills configured in the current workspace directory. This bug predates version 18 and has never worked for my setup. Now that we can place more than the copilot-instructions.md into the workspace folder, I would really like to use this feature.

**To Reproduce**

1. Set up a workspace with ABAP ADT installed and add a project to it
2. Create the .github directory beside the .metadata folder
3. Create the prompts and skills directories and fill them with .md files as described in the documentation. In the case of prompts, I created the file example.prompt.md
4. Type / in the chat box to see loaded prompts. Restarting Eclipse does not solve the issue.

**Expected behavior**
At least prompts should show up under /, as there is no obvious reason for them not to work in an ADT project. I have noticed that the .github folder is actually opened by Eclipse, so I am not sure what is stopping it from loading the prompts. I also took a look at the logs and found no errors regarding this issue.

Arexoor:
Fixes #245 

This PR fixes the issue where prompts and skills files located at the workspace root directory (e.g., eclipse-workspace/.github) were not being discovered and loaded by the Copilot plugin with an SAP ADT projects.

## Root Cause
The WorkspaceUtils.listWorkspaceFolders() method only returned top-level Eclipse projects in the workspace, but never included the workspace root directory itself. This meant that:

- Prompt files (.prompt.md) and skill files (SKILL.md) placed at the workspace root level under .github/ were never discovered
- The language server had no way to search the workspace root for these files
- In SAP ADT environments, this was particularly problematic as projects are not structured as standard Eclipse projects

## Solution
Modified WorkspaceUtils.listWorkspaceFolders() to:

- Include workspace root as the first workspace folder in the list
- Add deduplication logic to prevent duplicate URIs when a project is located at the workspace root
- Gracefully handle errors when accessing the workspace root URI

## Changes Made:
`com.microsoft.copilot.eclipse.core/src/com/microsoft/copilot/eclipse/core/utils/WorkspaceUtils.java`

- Modified listWorkspaceFolders() to include workspace root URI with name "workspace-root"
- Added a Set<String> for tracking seen URIs to prevent duplicates
- Added error handling for edge cases when retrieving workspace root

`com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/services/ChatCompletionService.java`

- Added debug logging to show which workspace folders are being used for prompt and skill discovery
- Logs the name and URI of each workspace folder being searched

### DIFF
--- a/com.microsoft.copilot.eclipse.core/src/com/microsoft/copilot/eclipse/core/utils/WorkspaceUtils.java
+++ b/com.microsoft.copilot.eclipse.core/src/com/microsoft/copilot/eclipse/core/utils/WorkspaceUtils.java
@@ -82,19 +82,42 @@ public class WorkspaceUtils {
   }
 
   /**
-   * List all top level projects as workspace folders in the current workspace.
+   * List all top level projects as workspace folders in the current workspace, including the workspace root.
    */
   public static List<WorkspaceFolder> listWorkspaceFolders() {
     List<IProject> projects = WorkspaceUtils.listTopLevelProjects();
 
     List<WorkspaceFolder> folders = new ArrayList<>();
+    java.util.Set<String> seenUris = new java.util.HashSet<>();
+
+    // Add workspace root first so it can be searched for root-level prompt files and skills
+    try {
+      URI workspaceRootUri = ResourcesPlugin.getWorkspace().getRoot().getLocationURI();
+      if (workspaceRootUri != null) {
+        String rootUriString = workspaceRootUri.toASCIIString();
+        WorkspaceFolder rootFolder = new WorkspaceFolder();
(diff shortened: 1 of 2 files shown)


iloveeclipse:
See also https://github.com/microsoft/copilot-for-eclipse/issues/143 for a more systematic problem we have. probably your fix need to be alighned with that?

PR #272, opened 2026-06-01
fix -  file and directory link navigation in chat tool results

### LINKED ISSUE #262: [Bug] Cannot navigate to the file out of workspace
1. Trigger a file based tool call for file out of the workspace
2. Click the link

Image

xinyi-gong:
Fixes #262

## Problem
Clicking file path links in Chat tool results fails to open files that are outside the Eclipse workspace. Directory links in the workspace also had no meaningful behavior.

## Changes

- When a file link does not resolve to a workspace resource, fall back to opening it as a local filesystem file in an editor.
- Handle workspace directory and project links by revealing them in the Project Explorer instead of opening the browser.
- Centralized local path resolution logic into the core bundle, removing duplication from the file tools.

### DIFF
--- a/com.microsoft.copilot.eclipse.core/src/com/microsoft/copilot/eclipse/core/utils/FileUtils.java
+++ b/com.microsoft.copilot.eclipse.core/src/com/microsoft/copilot/eclipse/core/utils/FileUtils.java
@@ -194,16 +194,29 @@ public class FileUtils {
     }
 
     try {
-      String pathWithoutFragment = stripFragment(filePath);
-      if (pathWithoutFragment.startsWith("file:")) {
-        return Paths.get(new URI(pathWithoutFragment)).toAbsolutePath().normalize();
+      // For file URIs, '#' is always a fragment delimiter (literal '#' in filenames is encoded as %23).
+      if (filePath.startsWith("file:")) {
+        String uriWithoutFragment = stripFragment(filePath);
+        return Paths.get(new URI(uriWithoutFragment)).toAbsolutePath().normalize();
       }
+
+      String pathWithoutFragment = stripFragment(filePath);
       if (URI_SCHEME_PATTERN.matcher(pathWithoutFragment).find() && !hasDriveLetter(pathWithoutFragment)) {
         return null;
       }
 
-      Path path = Paths.get(pathWithoutFragment);
-      return path.isAbsolute() ? path.toAbsolutePath().normalize() : null;
+      // For raw paths, try the full string first since '#' is a valid filename character on Unix/Linux.
+      // Only fall back to stripping the fragment if the full path doesn't exist.
+      Path fullPath = Paths.get(filePath);
+      if (fullPath.isAbsolute()) {
+        if (Files.exists(fullPath)) {
+          return fullPath.toAbsolutePath().normalize();
+        }
+        // Fall back: treat '#...' as a line-number fragment
+        Path strippedPath = Paths.get(pathWithoutFragment);
+        return strippedPath.isAbsolute() ? strippedPath.toAbsolutePath().normalize() : null;
+      }
+      return null;
     } catch (IllegalArgumentException | URISyntaxException e) {
       CopilotCore.LOGGER.error("Invalid local file path: " + filePath, e);
       return null;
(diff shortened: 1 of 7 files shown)


com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/FileAnnotationHyperlinkDetector.java
```diff
@@ -59,14 +62,27 @@ public void open() {
...
+        }
+        Path localPath = FileUtils.getLocalFilePath(urlString);
+        if (localPath != null && Files.isRegularFile(localPath, LinkOption.NOFOLLOW_LINKS)) {
+          UiUtils.openLocalFileInEditor(localPath);
           return;
         }
       } else {
-        IFile file = ResourcesPlugin.getWorkspace().getRoot().getFile(new Path(urlString));
+        IFile file = ResourcesPlugin.getWorkspace().getRoot()
+            .getFile(new org.eclipse.core.runtime.Path(urlString));
```
ethanyhou, line 85: Avoid FQN here

com.microsoft.copilot.eclipse.core/src/com/microsoft/copilot/eclipse/core/utils/FileUtils.java
```diff
@@ -181,6 +181,40 @@ public static IFile getFileFromUri(String fileUri) {
...
+      Path path = Paths.get(pathWithoutFragment);
+      return path.isAbsolute() ? path.toAbsolutePath().normalize() : null;
+    } catch (IllegalArgumentException | URISyntaxException e) {
+      CopilotCore.LOGGER.error("Invalid local file path: " + filePath, e);
+      return null;
+    }
+  }
+
+  private static String stripFragment(String pathOrUri) {
+    int fragmentIndex = pathOrUri.indexOf('#');
```
ethanyhou, line 227: Unix and Linux allow # in file paths. Could this introduce any risks?

xinyi-gong: For file URIs this is safe,  RFC 3986 requires literal # in filenames to be encoded as %23, so stripFragment won't misinterpret it. For raw filesystem paths, I've updated the logic to try the full path (including #) first; only if that path doesn't exist on disk does it fall back to treating #... as a line-number fragment.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/FileAnnotationHyperlinkDetector.java
xinyi-gong: There is already a java.nio.file.Path (see L78), so I have to user FQN here

com.microsoft.copilot.eclipse.core/src/com/microsoft/copilot/eclipse/core/utils/FileUtils.java
```diff
@@ -181,6 +181,53 @@ public static IFile getFileFromUri(String fileUri) {
...
+
+    try {
+      // For file URIs, '#' is always a fragment delimiter (literal '#' in filenames is encoded as %23).
+      if (filePath.startsWith("file:")) {
+        String uriWithoutFragment = stripFragment(filePath);
+        return Paths.get(new URI(uriWithoutFragment)).toAbsolutePath().normalize();
+      }
+
+      String pathWithoutFragment = stripFragment(filePath);
+      if (URI_SCHEME_PATTERN.matcher(pathWithoutFragment).find() && !hasDriveLetter(pathWithoutFragment)) {
```
jdneo, line 204: Question: why we restrict that the uri cannot have drive letter?

xinyi-gong: Without the hasDriveLetter exclusion, Windows absolute paths like C:/Users/file.txt would be incorrectly rejected as non-local URIs. I only want to filter non-local URIs here.(like https://).

jdneo (approved):

PR #280, opened 2026-06-05
fix: clear closed editor file context

### LINKED ISSUE #277: The Copilot Chat view retains file context even after the file is closed
**Environment**
- OS: windows 11
- Eclipse Version: 4.34
- Plugin Version: 0.18.0

**Describe the bug**
The Copilot Chat view retains file context even after the file is closed

**To Reproduce**

1 Ensure no file is open in Eclipse. Ensure the active view is not Copilot Chat (e.g., keep the Problems view active).
2 Open a single file in the editor.
3 Switch the active view to Copilot Chat.
4 Close the file.

**Expected behavior**
The file context should be cleared immediately after the file is closed, regardless of when the Copilot Chat view was activated.

**Additional context**
If the Copilot Chat view is active before opening the file, then the context is correctly cleared upon file closure.
This indicates inconsistent behavior based on view activation timing.

nanookclaw:
## Summary

Fixes #277.

When an editor closes, `ReferencedFileService` now checks whether the closed editor is the file currently represented in Copilot Chat. If it is, the service clears both the current-file observable and current-selection observable immediately. This keeps the existing fallback behavior for the final-editor-close case, but no longer depends only on the workbench editor reference count, which can still include an editor while focus has moved to a non-editor view.

The regression test captures the part listener registered by `ReferencedFileService`, seeds the current file observable, simulates closing that same editor while another editor reference is still reported by the active page, and asserts the chat current file is cleared.

## Testing

- `git diff --check`
- `./mvnw -pl com.microsoft.copilot.eclipse.terminal.api,com.microsoft.copilot.eclipse.core,com.microsoft.copilot.eclipse.ui,com.microsoft.copilot.eclipse.ui.test -am -Dtest=ReferencedFileServiceTest test`

I also tried the same module set with `verify`, but it stops in packaging because `com.microsoft.copilot.eclipse.core/build.properties` expects `copilot-agent/dist/`, which is not present in this checkout.

### DIFF
--- a/com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/services/ReferencedFileService.java
+++ b/com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/services/ReferencedFileService.java
@@ -26,13 +26,19 @@ import org.eclipse.jface.viewers.SelectionChangedEvent;
 import org.eclipse.lsp4e.LSPEclipseUtils;
 import org.eclipse.lsp4j.Position;
 import org.eclipse.lsp4j.Range;
+import org.eclipse.ui.IEditorInput;
 import org.eclipse.ui.IEditorPart;
+import org.eclipse.ui.IEditorReference;
+import org.eclipse.ui.IFileEditorInput;
 import org.eclipse.ui.IPartListener2;
+import org.eclipse.ui.IWorkbench;
 import org.eclipse.ui.IWorkbenchPage;
 import org.eclipse.ui.IWorkbenchPart;
 import org.eclipse.ui.IWorkbenchPartReference;
 import org.eclipse.ui.IWorkbenchWindow;
+import org.eclipse.ui.PartInitException;
 import org.eclipse.ui.PlatformUI;
+import org.eclipse.ui.ide.ResourceUtil;
 import org.eclipse.ui.texteditor.ITextEditor;
 
 import com.microsoft.copilot.eclipse.core.Constants;
@@ -478,12 +484,7 @@ public class ReferencedFileService extends ChatBaseService implements IReference
   }
 
   private boolean isCurrentReferencedFile(IWorkbenchPartReference partRef) {
-    IWorkbenchPart part = partRef.getPart(false);
-    if (!(part instanceof IEditorPart editorPart)) {
-      return false;
-    }
-
-    IFile closedFile = UiUtils.getFileFromEditorPart(editorPart);
+    IFile closedFile = getFileFromPartReference(partRef);
     if (closedFile == null) {
       return false;
     }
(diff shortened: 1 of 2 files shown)


jdneo:
Hi @nanookclaw

Thank you for your contribution, there are some copilot review comments, could you first check if they are valid?

nanookclaw:
Checked the Copilot review comments and they were valid. I pushed `122edf0` with the fixes:

- Resolve the closed-editor false negative by reading the file from `IEditorReference.getEditorInput()` before falling back to the editor part.
- Avoid active-page-only clearing by checking editor references across all workbench windows/pages.
- Replace the test reflection against `currentFileObservable` with a small protected test seam and local test subclass.

Verification:
- `git diff --check`
- `./mvnw -pl com.microsoft.copilot.eclipse.core,com.microsoft.copilot.eclipse.terminal.api,com.microsoft.copilot.eclipse.ui,com.microsoft.copilot.eclipse.ui.test -Dtest=ReferencedFileServiceTest test`

jdneo:
Thank you. Please mark those comments as resolved once you addressed them

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/services/ReferencedFileService.java
```diff
@@ -105,12 +111,10 @@ public void partBroughtToTop(IWorkbenchPartReference partRef) {
...
       @Override
       public void partClosed(IWorkbenchPartReference partRef) {
-        IWorkbenchPage page = UiUtils.getActivePage();
-        if (page == null || page.getEditorReferences().length == 0) {
-          ensureRealm(() -> {
-            currentFileObservable.setValue(null);
-            currentSelectionObservable.setValue(null);
-          });
+        boolean closedCurrentFile = isCurrentReferencedFile(partRef);
+        boolean noOpenEditors = hasNoOpenEditors();
```
jdneo, line 115: Curious about the no open editors check here.

From code, seems like `closedCurrentFile` is enough, why we have this check?

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/services/ReferencedFileService.java
```diff
@@ -479,6 +483,98 @@ private void updateCurrentReferencedFile(IWorkbenchPartReference partRef) {
...
+    ensureRealm(() -> {
+      currentFileObservable.setValue(null);
+      currentSelectionObservable.setValue(null);
+    });
+  }
+
+  /**
+   * Sets the current file for tests.
+   */
+  protected void setCurrentFileForTest(IFile file) {
```
jdneo, line 574: Instead of introduce a method for testing purpose only, can we achieve the same purpose by firing a part-open/activate event on the captured listener before firing partClosed in test?

jdneo (commented):
Hi @nanookclaw

Thank you for the contribution and sorry for the late reply. Left some comments.

PR #299, opened 2026-06-21
Forward MCP streamable HTTP headers

### LINKED ISSUE #295: [Bug] Custom HTTP headers from MCP config are not sent on Streamable HTTP requests
## Description

When configuring a remote MCP server with custom `headers` in the MCP JSON config (either via Eclipse), the custom headers are **not included** in the Streamable HTTP POST requests sent to the server.

Only standard HTTP headers are sent (`host`, `accept`, `content-type`, `mcp-protocol-version`, `mcp-session-id`, `user-agent`, etc.), but any user-defined headers (e.g., `X-Username`, `X-Api-Token`) are silently dropped. 

## Steps to Reproduce

1. Configure a remote MCP server in Eclipse with type `http` and custom headers:
```json
{
  "servers": {
    "my-server": {
      "type": "http",
      "url": "https://my-mcp-server.example.com/mcp",
      "headers": {
        "X-Username": "myuser",
        "X-Api-Token": "my-token"
      }
    }
  }
}
```
2. Open a Copilot chat and invoke a tool from the configured MCP server.
3. On the server side, inspect the incoming HTTP request headers.

## Expected Behavior

The custom headers (`X-Username`, `X-Api-Token`) should be included in every HTTP POST request to the MCP server endpoint, as specified in the [MCP specification for Streamable HTTP transport](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#streamable-http).

## Actual Behavior

Only standard HTTP headers are present. The custom headers from the config are silently ignored. Server-side log output shows:

```
header_keys=['x-forwarded-for', 'x-request-id', 'host', 'accept', 'content-type', 'mcp-protocol-version', 'mcp-session-id', 'user-agent', ...]
```

No custom headers are included.

## Workaround

As a workaround, authentication tokens can be passed via URL query parameters instead of headers:
```json
{
  "servers": {
    "my-server": {
      "type": "http",
      "url": "https://my-mcp-server.example.com/mcp?username=myuser&api_token=my-token"
    }
  }
}
```
However, this is less secure since tokens appear in server access logs and URL histories.

## Environment

- Eclipse: 2025-12 (4.38)
- GitHub Copilot for Eclipse:   GitHub Copilot	0.18.0.202606020012_nightly	com.microsoft.copilot.eclipse.feature.feature.group	GitHub Copilot
- OS: Linux (RHEL 9)

## Notes

This works correctly in VS Code and the GitHub Copilot CLI, where custom headers from MCP config are properly forwarded. The issue appears to be specific to the Eclipse plugin's Streamable HTTP client implementation.

jstar0:
Fixes #295

## Summary

Forwards custom MCP server HTTP headers from the top-level `headers` field into `requestInit.headers`.

## Changes

This PR:
- copies each MCP server's top-level `headers` into `requestInit.headers`
- keeps the original top-level `headers` field intact
- preserves any existing `requestInit.headers` value when the same header is already set there
- supports both VS Code-style `{ "servers": ... }` config and direct server-map config
- skips malformed non-object `requestInit` or `requestInit.headers` shapes instead of replacing user-provided values
- adds regression coverage for the wrapped config, direct config, and duplicate-header precedence cases

## Why

Streamable HTTP MCP servers may require custom HTTP headers for authentication. The settings adapter already forwards MCP server configuration to the language server, but top-level `headers` were not represented in `requestInit.headers`, so those headers were not available to the Streamable HTTP request path.

## Verification

```bash
./mvnw -pl com.microsoft.copilot.eclipse.core,com.microsoft.copilot.eclipse.core.test -am -Dtest=CopilotLanguageServerSettingsTests verify
```

### DIFF
--- a/com.microsoft.copilot.eclipse.core.test/src/com/microsoft/copilot/eclipse/core/lsp/protocol/CopilotLanguageServerSettingsTests.java
+++ b/com.microsoft.copilot.eclipse.core.test/src/com/microsoft/copilot/eclipse/core/lsp/protocol/CopilotLanguageServerSettingsTests.java
@@ -1,0 +1,103 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+package com.microsoft.copilot.eclipse.core.lsp.protocol;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import com.google.gson.Gson;
+import com.google.gson.JsonObject;
+import org.junit.jupiter.api.Test;
+
+class CopilotLanguageServerSettingsTests {
+
+  private final Gson gson = new Gson();
+
+  @Test
+  void testSetMcpServers_copiesTopLevelHeadersToRequestInitHeaders() {
+    CopilotLanguageServerSettings settings = new CopilotLanguageServerSettings();
+    String preference = """
+        {
+          "servers": {
+            "remote": {
+              "type": "http",
+              "url": "https://example.com/mcp",
+              "headers": {
+                "X-Username": "myuser",
+                "X-Api-Token": "my-token"
+              }
+            }
+          }
+        }
+        """;
+
+    settings.setMcpServers(preference);
+
+    JsonObject mcpServers = gson.fromJson(settings.getGithubSettings().getCopilotSettings().getMcpServers(),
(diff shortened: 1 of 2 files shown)


jdneo:
Thank you for your contribution. This should already work in 0.19.0. So no change is needed.

If it is wrong, please leave comments, thank you.

PR #316, opened 2026-06-23
fix: Updated token pricing model with context size and tier details

ethanyhou:
Align context size calculation algorithm with Intellij:
image
image

### DIFF
--- a/com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ActionBar.java
+++ b/com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ActionBar.java
@@ -580,7 +580,6 @@ public class ActionBar extends Composite implements NewConversationListener {
   }
 
   private void setUpContextSizeDonutInControlBar(Composite parent) {
-    chatServiceManager.getContextWindowService().setModelService(chatServiceManager.getModelService());
     this.contextSizeDonut = new ContextSizeDonut(parent, chatServiceManager.getContextWindowService());
   }
 
--- a/com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ActionBar.java
+++ b/com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ActionBar.java
@@ -580,6 +580,7 @@ public class ActionBar extends Composite implements NewConversationListener {
   }
 
   private void setUpContextSizeDonutInControlBar(Composite parent) {
+    chatServiceManager.getContextWindowService().setModelService(chatServiceManager.getModelService());
     this.contextSizeDonut = new ContextSizeDonut(parent, chatServiceManager.getContextWindowService());
   }
 
--- a/com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/utils/ModelUtils.java
+++ b/com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/utils/ModelUtils.java
@@ -163,7 +163,7 @@ public class ModelUtils {
   /**
    * Returns the formatted context window size for the model, or {@code null} if unavailable.
    */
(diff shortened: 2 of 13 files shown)


com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/utils/ModelUtils.java
```diff
@@ -162,15 +163,60 @@ public static String formatPriceCategory(String priceCategory) {
...
+      if (maxInputTokens != null && maxOutputTokens != null) {
+        return maxInputTokens + maxOutputTokens;
+      }
+    }
+
+    return limits.maxContextWindowTokens();
+  }
+
+  private static boolean isTokenBasedBillingEnabled(CopilotModel model) {
+    return model.getBilling() != null && model.getBilling().tokenBasedBillingEnabled();
```
jdneo, line 210: Could you help me to hint my memory: when will a model's TBB disabled?

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/utils/ModelUtils.java
```diff
@@ -162,15 +163,60 @@ public static String formatPriceCategory(String priceCategory) {
...
+    CopilotModelCapabilitiesLimits limits = model.getCapabilities().limits();
+    Integer maxOutputTokens = limits.maxOutputTokens();
+    int output = maxOutputTokens == null ? 0 : maxOutputTokens;
+
+    CopilotModelTokenPriceTier defaultTier = getDefaultTokenPriceTier(model);
+    if (defaultTier != null && defaultTier.maxContext() != null) {
+      return defaultTier.maxContext() + output;
+    }
+
+    if (isTokenBasedBillingEnabled(model)) {
```
jdneo, line 198: So according to the code, if a model does not have a default tier, logic falls into here. Is it only for defensive purpose?

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ActionBar.java
```diff
@@ -580,6 +580,7 @@ public void widgetSelected(SelectionEvent e) {
   }
 
   private void setUpContextSizeDonutInControlBar(Composite parent) {
+    chatServiceManager.getContextWindowService().setModelService(chatServiceManager.getModelService());
```
jdneo: My two cents here:

Checking the ChatServiceManager, the ModelService is initialized earlier than ContextWindowService. That means you can inject the ModelService into ContextWindowService via constructor, that avoids introducing the setter here.

Though this will require the model service be initialized earlier than ContextWindowService, but that is not a big problem and will be more natural if we introduce DI into our project in future.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/utils/ModelUtils.java
ethanyhou: Legacy paid annual users may still have TBB not enabled accounts. Added TODO as other places to remind us to clean the code in the future.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/utils/ModelUtils.java
ethanyhou: Yes, this is only for defensive purpose. Also added TODO as other places to remind us to clean the code in the future.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ActionBar.java
ethanyhou: Good point, updated.

jdneo (approved):

PR #321, opened 2026-06-25
fix - reduce streaming render cost with incremental layout and event batching

### LINKED ISSUE #259: [Bug] After updating to 0.18.0 Agents are extremely slow
After the latest updates, Agents are extremely slow, thinking for extremely long time, even the moderate ones. Tasks that used to take 10-15 minutes, now takes several hours. Mostly tested with GPT 5.4-medium.

xinyi-gong:
fix https://github.com/microsoft/copilot-for-eclipse/issues/259

## Root cause

Every streaming chunk (text fragment, thinking update, tool call delta) triggered a **full**
`layout(true, true)` + `cmpContent.computeSize(SWT.DEFAULT, SWT.DEFAULT)` pass that re-measured
**every historical turn**. 

### 1. Event-drain batching (`drainPendingEvents`)
Streaming events are now queued into a `ConcurrentLinkedQueue` and drained in a single scheduled
UI-thread callback instead of dispatching one layout pass per event. A re-entrancy guard prevents
cascading drains from `controlResized`.

### 2. Incremental / O(1) layout
During streaming only the **trailing turns** (latest user turn + active Copilot turn) are
re-measured. Historical sealed turns keep their cached sizes (`computeSize(w, DEFAULT, false)`).
This replaces the previous `layout(true, true)` call that invalidated the entire composite.

### 3. Constrained-width size measurement (flicker fix)
`computeSize` is now called with the actual `clientArea.width` instead of `SWT.DEFAULT`.
Previously the unconstrained width produced a phantom over-estimated height, causing the scroll
position to oscillate during streaming and producing visible flicker.

### 4. Drop per-chunk enclosing-scroller refresh in `ThinkingBlock` streaming
The thinking streaming path no longer calls `refreshEnclosingScroller()` on every chunk. Long
thinking content previously forced a full enclosing `ChatContentViewer` re-layout per chunk
(hundreds of full O(n) passes per turn). The enclosing scroller is now refreshed by the batched
`drainPendingEvents` → `doRefreshScrollerLayout` path along with all other streaming events; the
local `updateScrollerDuringStreaming()` only manages the thinking block's own inner scroller. The
expand/collapse path still refreshes the enclosing scroller directly.

### 5. Incremental re-measure on turn-end only
The drain batch triggers a incremental re-measure only when it contains a `turn-end` event, ensuring
`minHeight` is accurate before `scrollToBottom()`. 

## Performance results

Benchmark: deterministic repeated-Ask prompt, 5 turns (`n` = widget count, increments by 2 per turn).
Compared against baseline run on unfixed HEAD.

| turn | refresh avg (ms) base → fix | refresh total (ms) base → fix | render % of wall-clock base → fix |
|------|-----------------------------|---------------------------------|-----------------------------------|
| n2   | 20.04 → 5.49                | 13 650 → 3 168                  | 33% → 10%                         |
| n4   | 34.38 → 5.34                | 23 960 → 2 628                  | 55% → 9%                          |
| n6   | 46.79 → 9.01                | 38 414 → 5 787                  | 68% → 15%                         |
| n8   | 52.19 → 6.58                | 42 011 → 4 329                  | 70% → 11%                         |
| n10  | 62.47 → 9.56                | **49 910 → 6 464**              | **77% → 17%**                     |

**~7.7× reduction in layout time at turn 5.** The per-pass average is now essentially flat
(5.5–9.6 ms) across n2–n10 — effectively O(1) in turn count. Per turn only 2–4 full-measure
passes run; the rest take the O(1) incremental path. Render is no longer the dominant cost.

Perf Doc: [perf-baseline-259.md](https://github.com/user-attachments/files/29367640/perf-baseline-259.md)

## Files changed
- `ChatContentViewer.java` — drain queue, incremental layout, re-entrancy guard, flicker fix,
  full-measure-on-turn-end, auto-scroll logic
- `ThinkingBlock.java` — parametrised `refreshEnclosingScroller(boolean incremental)`
- `BaseTurnWidget.java` — minor cleanup
- `ChatMarkupViewer.java` — minor cleanup

### DIFF
--- a/com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatContentViewer.java
+++ b/com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatContentViewer.java
@@ -83,17 +83,8 @@ public class ChatContentViewer extends ScrolledComposite {
   /** Client-area width of the last layout pass; a change forces a full re-measure (text re-wraps). */
   private int lastLayoutWidth = -1;
 
-  /**
-   * Guards {@link #doRefreshScrollerLayout(boolean)} against synchronous re-entrancy: setting the
-   * scroller min size can toggle the scrollbar, changing the client width and re-firing
-   * {@code controlResized}. The running pass records any nested request here and drains it after
-   * unwinding (a full re-measure wins over an incremental one).
-   */
-  private enum LayoutState {
-    IDLE, RUNNING, RERUN_INCREMENTAL, RERUN_FULL
-  }
-
-  private LayoutState layoutState = LayoutState.IDLE;
+  /** Guards against scheduling more than one pending async refresh from a burst of resize events. */
+  private final AtomicBoolean refreshScheduled = new AtomicBoolean(false);
 
   /**
    * Create the composite.
@@ -120,7 +111,7 @@ public class ChatContentViewer extends ScrolledComposite {
     this.addControlListener(new ControlAdapter() {
       @Override
       public void controlResized(ControlEvent e) {
-        refreshScrollerLayout();
+        scheduleCoalescedRefresh();
       }
     });
 
@@ -224,21 +215,31 @@ public class ChatContentViewer extends ScrolledComposite {
       pendingEvents.clear();
       return;
     }
-    boolean sawTurnEnd = false;
     ChatProgressValue event;
+    boolean sawTurnEnd = false;
     while ((event = pendingEvents.poll()) != null) {
-      final ChatProgressValue e = event;
-      if (e.getKind() == WorkDoneProgressKind.end) {
+      doProcessTurnEvent(event);
+      if (event.getKind() == WorkDoneProgressKind.end) {
         sawTurnEnd = true;
       }
-      doProcessTurnEvent(e);
     }
-    // Use a full re-measure on the turn-end batch so minHeight is accurate before scrollToBottom();
-    // the incremental path can accumulate a stale containerSize that leaves phantom space below the
-    // real content. All streaming batches keep the O(1) incremental path.
-    doRefreshScrollerLayout(sawTurnEnd);
+    refreshScrollerLayout(false);
     if (shouldAutoScrollToBottom()) {
       scrollToBottom();
+    }
+    if (sawTurnEnd) {
+      // A turn's height settles one frame after its content changes. Mid-stream the next chunk's drain
+      // re-measures and re-scrolls into that settled height; the final chunk has no follow-up, so run
+      // one on the next frame.
+      SwtUtils.invokeOnDisplayThreadAsync(() -> {
+        if (isDisposed()) {
+          return;
+        }
+        refreshScrollerLayout(false);
+        if (shouldAutoScrollToBottom()) {
+          scrollToBottom();
+        }
+      }, this);
     }
     // Events may have arrived while draining; schedule a follow-up drain if so.
     if (!pendingEvents.isEmpty() && drainScheduled.compareAndSet(false, true)) {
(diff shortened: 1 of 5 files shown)


com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatContentViewer.java
```diff
@@ -437,11 +489,18 @@ public void renderErrorMessage(String errorMessage) {
...
+   * turn outside the streaming event drain.
    */
-  public void requestRefreshScrollerLayout() {
-    SwtUtils.invokeOnDisplayThreadAsync(() -> refreshScrollerLayout(), this);
+  public void scheduleRefresh() {
+    if (this.isDisposed()) {
+      return;
+    }
+    SwtUtils.invokeOnDisplayThreadAsync(() -> {
+      if (!this.isDisposed()) {
```
jdneo: Since here we checked `isDisposed()`, the outer if check can be removed.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatContentViewer.java
jdneo: This function looks very similar with scheduleRefresh(), consider only leave one of them?

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatContentViewer.java
```diff
@@ -186,123 +208,153 @@ public void setConversationId(String conversationId) {
...
+    if (isDisposed()) {
+      pendingEvents.clear();
+      return;
+    }
+    boolean sawTurnEnd = false;
+    ChatProgressValue event;
+    while ((event = pendingEvents.poll()) != null) {
+      final ChatProgressValue e = event;
+      if (e.getKind() == WorkDoneProgressKind.end) {
+        sawTurnEnd = true;
```
jdneo: What will happen if we also only do incremental pass when turn ends? Since the previous turns' size won't change?

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatContentViewer.java
```diff
@@ -451,32 +510,119 @@ public void refreshScrollerLayout() {
     if (this.isDisposed()) {
       return;
     }
+    // Public entry point for structural changes (turn start, error/warn widgets, expand/collapse of
+    // historical thinking blocks). These can resize a non-trailing turn, so force a full re-measure.
+    // The streaming path instead goes through the event drain, which stays O(1).
+    doRefreshScrollerLayout(true);
+  }
+
+  private void doRefreshScrollerLayout(boolean forceFullMeasure) {
```
jdneo: doRefreshScrollerLayoutDecide, doRefreshScrollerLayoutApply, and the LayoutState enum make the layout path harder to follow than necessary. Could we collapse the decide/apply methods into a single refreshScrollerLayout(boolean forceFullMeasure) and avoid synchronous re-entrancy by making controlResized schedule a coalesced async refresh instead of calling the layout method directly? That should preserve the full-vs-incremental behavior while removing the state machine.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatContentViewer.java
xinyi-gong: You're right. Updated to an incremental pass. Full re-measure is now reserved for structural changes that can resize a non-trailing turn (turn start, error/warn widgets, expand/collapse of historical thinking blocks).

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatContentViewer.java
xinyi-gong: Updated

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatContentViewer.java
xinyi-gong: Removed

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatContentViewer.java
xinyi-gong: Removed

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatContentViewer.java
```diff
@@ -186,123 +199,163 @@ public void setConversationId(String conversationId) {
...
+      doProcessTurnEvent(event);
+      if (event.getKind() == WorkDoneProgressKind.end) {
+        sawTurnEnd = true;
+      }
+    }
+    refreshScrollerLayout(false);
+    if (shouldAutoScrollToBottom()) {
+      scrollToBottom();
+    }
+    if (sawTurnEnd) {
```
jdneo: Why we need to schedule another

```
refreshScrollerLayout(false);
        if (shouldAutoScrollToBottom()) {
          scrollToBottom();
        }
```

when saw turn end?

xinyi-gong: I removed this extra turn-end refresh.

After looking into it more, the `end` event is not a reliable point where the UI layout is already stable. Handling `end` can flush buffered markdown into the turn, and some child viewers may still update their measured height afterwards. So doing one more refresh specifically when seeing `end` is still timing-based and does not fully address the root cause.

Instead, the updated approach makes auto-scroll preservation explicit on refresh calls that are part of active-turn UI updates. For example, the model-info/footer refresh that happens after the response is finalized now preserves auto-scroll, so the refresh that actually observes the final height also keeps the view at the bottom if the user has not scrolled away. Generic resize/coalesced refreshes do not preserve auto-scroll, to avoid pulling the user back down unexpectedly.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatContentViewer.java
```diff
@@ -186,123 +199,143 @@ public void setConversationId(String conversationId) {
...
+    drainScheduled.set(false);
+    if (isDisposed()) {
+      pendingEvents.clear();
+      return;
+    }
+    ChatProgressValue event;
+    while ((event = pendingEvents.poll()) != null) {
+      doProcessTurnEvent(event);
+    }
+    refreshScrollerLayoutInternal(false);
```
jdneo: The three "refresh layout" methods here are hard to tell apart because their names describe *how/when they're scheduled* rather than *what they do*. The only real difference between them is the measurement mode (full vs. incremental), and that's hidden behind a bare boolean — at the call sites `refreshScrollerLayoutInternal(false)` / `(true)` give no hint what the argument means.

Could we make the mode explicit and drop the boolean? Two intent-revealing methods over one shared impl reads much better:

```java
/** Full re-measure of all turns. Layout only; callers scroll separately. */
public void refreshLayoutFull() {
  refreshLayout(MeasureMode.FULL);
}

/** Incremental re-measure of just the trailing (streaming) turns. */
private void refreshLayoutIncremental() {
  refreshLayout(MeasureMode.INCREMENTAL);
}

/** Coalesces a burst of resize events into a single async incremental pass. */
private void scheduleIncrementalRefresh() {
  if (refreshScheduled.compareAndSet(false, true)) {
    SwtUtils.invokeOnDisplayThreadAsync(() -> {
      refreshScheduled.set(false);
      refreshLayoutIncremental();
    }, this);
  }
}

private enum MeasureMode { FULL, INCREMENTAL }

private void refreshLayout(MeasureMode mode) {
  if (this.isDisposed()) {
    return;
  }
  // A width change always forces FULL because soft-wrapped text re-wraps.
  boolean fullMeasure = mode == MeasureMode.FULL || width != lastLayoutWidth;
  ...
}
```

Call sites then explain themselves: `refreshLayoutIncremental()` in `drainPendingEvents`, `scheduleIncrementalRefresh()` in `controlResized`, and `refreshLayoutFull()` in `startNewTurn` / `renderErrorMessage` / the compacting handlers. `FULL` / `INCREMENTAL` also map directly to the O(n)-vs-O(1) model described in the PR.

If you'd rather keep the public `refreshScrollerLayout()` name for external callers, we could leave it as a thin alias to `refreshLayoutFull()` — but at minimum I'd love to see the `(true)`/`(false)` boolean replaced with the enum so the call sites are readable.

xinyi-gong: Updated. Thanks for your suggestion.

jdneo (approved):
Solid fix, the performance is much better than before.

Thanks.

PR #323, opened 2026-06-25
Fix visible non-workspace editor context for chat

### LINKED ISSUE #117: Unable to use the Java source file corresponding to the class file as context
Hello, the platform is: Eclipse 4.39 + Copilot 0.16.0.

When importing a plugin project, if the imported project only contains JAR files and does not include source code, Eclipse can open the class file and display the source files associated with that JAR. However, in Copilot, the source file corresponding to that class file cannot be selected as the AI's context. If Copilot is prompted at this point, it will search for the Java file, but the Java file is not in the project. The output will be:
...
Searched for files matching query: org.eclipse.core.resources/*/.java
Read directory org.eclipse.core.resources
...
In other words, the corresponding Java file cannot be found for analysis. Therefore, the AI ​​cannot read the Java file corresponding to the class file and cannot use it as context for analysis.

mralonsocorona:
## Summary

Fixes current-editor chat context for visible Eclipse text editors that expose an `IDocument` but are not backed by a workspace `IFile`, such as JDT class-file/source-attachment editors.

The existing workspace `IFile` path remains unchanged. When no workspace file is available, the active visible editor document is exposed through a read-only internal `copilot-visible-editor://...` URI, and `workspace/readFile` can resolve that URI through the referenced-file service.

Fixes #117.

## Details

- Keeps `IFile` current-editor context as the first-priority path.
- Adds URI-based current-editor fallback support to `IReferencedFileService`.
- Adds read-only visible-editor snapshot support for non-`IFile` text editors.
- Keeps `FileUtils.readFileWithStats(...)` as the primary `workspace/readFile` path.
- Falls back to the visible-editor snapshot only when the normal read returns the existing `file not found: <uri>` result.
- Does not add a JDT dependency, hidden workspace project, indexing, or writable file semantics.

## Validation

- `git diff --check`
- `./mvnw checkstyle:check`
- `./mvnw test`

### DIFF
--- a/com.microsoft.copilot.eclipse.core.test/src/com/microsoft/copilot/eclipse/core/lsp/CopilotLanguageClientTests.java
+++ b/com.microsoft.copilot.eclipse.core.test/src/com/microsoft/copilot/eclipse/core/lsp/CopilotLanguageClientTests.java
@@ -140,19 +140,24 @@ class CopilotLanguageClientTests {
   }
 
   @Test
-  void testReadFileFallsBackToCurrentEditorWhenFileNotFound() throws InterruptedException, ExecutionException {
+  void testReadFileFallsBackToCurrentEditorWhenWorkspaceFileMissing()
+      throws InterruptedException, ExecutionException {
     // Arrange
     String uri = "copilot-visible-editor://current/1";
+    ReadFileResult fileResult = new ReadFileResult("workspace read failed", null);
     ReadFileResult currentEditorResult = new ReadFileResult("class Example {}", null);
 
-    try (MockedStatic<CopilotCore> copilotCoreMock = Mockito.mockStatic(CopilotCore.class)) {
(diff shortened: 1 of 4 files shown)


mralonsocorona:

jdneo (commented):
Looks like this pr only handles for the current file, what about user manually attached .class files? Will that belong to this PR's scope?

mralonsocorona:
Thanks for raising this.

I intentionally kept this PR scoped to the current-editor path because that is the behavior described in #117: the user has a visible Java source/class-file editor, but chat cannot use that visible editor content as context.

Manually attached ".class" files seem related, but I think they are a separate path from the current-editor context. Supporting that properly would likely require extending the referenced-file attachment flow/content resolution as well, rather than only the current-editor fallback added here.

My preference would be to keep this PR focused on the visible current editor, and handle manually attached ".class" files in a follow-up PR if you agree.

jdneo:
Thank you. I'm fine with that.

PR #331, opened 2026-07-01
feat: Auto-approve reads of Copilot customization files

### LINKED ISSUE #318: [Bug] Copilot asks read permission for a global skill file
Image

xinyi-gong:
### Problem

Reading a Copilot customization file (skill, instruction, prompt, or agent) is normal agent operation, but these reads always prompted for confirmation when *"Auto approve file operations not covered by rules"* was unchecked:

- Workspace instruction files under `.github/instructions/*` are matched by the default deny rule, so their reads prompt.
- User-global files (`~/.copilot`, `~/.claude`, `~/.agents`) are outside the workspace, so they always prompt.
- Skill folders (`SKILL.md` plus the helper files it references) had no special handling.

### Solution

Recognize these reads precisely from the language server, which is the source of truth for where the files live (including user-configured and global locations) — instead of guessing conventional paths.

New `copilot/customSkill|customPrompt|customInstruction|customAgent/list` requests report each file's on-disk URI. A read is auto-approved when it **exactly matches a reported file** or **sits inside a reported skill folder**. Editing is never auto-approved — only reads.

### Changes

- **`CustomizationFileService`** (`core`): keeps an immutable snapshot of customization-file paths per type, refreshed on the matching `copilot/custom*/didChange` notification. Exposed via `IChatServiceManager`.
- **Protocol**: 4 `@JsonRequest` list methods on `CopilotLanguageServer` + connection wrappers, returning `CustomizationFileInfo` (carries the file `uri`).
- **`FileOperationConfirmationHandler`**: auto-approves a `FILE_READ` of a recognized customization file, before the outside-workspace check.
- **`FileUtils.isPathWithin`**: path-containment helper for skill-folder matching.

### Testing

- `FileUtilsTests` — `isPathWithin` (self / strict / nested / sibling / null).
- `FileOperationConfirmationHandlerTests` — `isCustomizationRead` (exact match, skill folder + helpers, negatives).
- Test plan `file-operation-auto-approve.md` — TC-010 (workspace + edit-still-prompts) and TC-011 (user-global).

`./mvnw verify` green across all modules.

fix https://github.com/microsoft/copilot-for-eclipse/issues/318


com.microsoft.copilot.eclipse.core/src/com/microsoft/copilot/eclipse/core/lsp/CopilotLanguageClient.java
```diff
@@ -277,14 +280,41 @@ public void onDidChangeCustomSkill(Object params) {
...
     notifyCustomizationFilesChanged();
+    refreshCustomizationFiles(CustomizationType.PROMPT);
   }
 
+  /**
+   * Notify when custom instructions change (global or workspace). Signal-only; clients re-fetch the file list.
+   */
+  @JsonNotification("copilot/customInstruction/didChange")
+  public void onDidChangeCustomInstruction(Object params) {
+    refreshCustomizationFiles(CustomizationType.INSTRUCTION);
```
jdneo: Is it possible to leverage notifyCustomizationFilesChanged here?

Using event bus to decouple the client.java and the service.java

// I assume some detailed data needs to be introduced in that event, let's discuss first if you buy in this.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/services/ChatServiceManager.java
```diff
@@ -51,6 +53,8 @@ public ChatServiceManager() {
     referencedFileService = new ReferencedFileService();
     mcpConfigService = new McpConfigService();
     mcpExtensionPointManager = new McpExtensionPointManager(mcpConfigService);
+    customizationFileService = new CustomizationFileService(this.lsConnection);
+    customizationFileService.refreshAll();
```
jdneo: Check if refreshall is triggered in ui thread? If yes, reschedule it to a backend job.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/confirmation/FileOperationConfirmationHandler.java
```diff
@@ -338,6 +352,39 @@ private boolean isOutsideWorkspace(InvokeClientToolConfirmationParams params) {
...
+   * @param skillFolders reported skill folders (each directory containing a {@code SKILL.md})
+   * @return {@code true} when the read is a recognized customization read
+   */
+  public static boolean isCustomizationRead(Path file, Set<Path> customizationFiles, Set<Path> skillFolders) {
+    Path normalized = file.toAbsolutePath().normalize();
+    if (customizationFiles.contains(normalized)) {
+      return true;
+    }
+    for (Path skillFolder : skillFolders) {
+      if (FileUtils.isPathWithin(skillFolder, normalized, false)) {
```
jdneo: This is the only one place using `isPathWithin`, and `strict` is `false`, in that case, could you just simply inline compare using some kind of api of Path?

com.microsoft.copilot.eclipse.core/src/com/microsoft/copilot/eclipse/core/lsp/CopilotLanguageClient.java
xinyi-gong: Updated to use event bus to trigger refreshing.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/services/ChatServiceManager.java
xinyi-gong: Updated.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/confirmation/FileOperationConfirmationHandler.java
xinyi-gong: Updated

com.microsoft.copilot.eclipse.core/src/com/microsoft/copilot/eclipse/core/chat/service/CustomizationFileService.java
```diff
@@ -0,0 +1,144 @@
...
+    return customizationFiles;
+  }
+
+  @Override
+  public Set<Path> getSkillFolders() {
+    return skillFolders;
+  }
+
+  @Override
+  public void refreshAll() {
```
jdneo: nit: Consider rename to refreshAllAsync to indicate it is an async call?

xinyi-gong: Renamed

jdneo (approved):
LGTM

PR #334, opened 2026-07-02
fix - chat not scrolling to the newest messages in long conversations

### LINKED ISSUE #63: At some point not possible to scroll chat down to see last messages
### Feedback source

None

### GitHub Copilot plugin version

0.14.0.202512112209

### Eclipse IDE and version

Eclipse IDE for Enterprise Java and Web Developers (includes Incubating components) - 2025-12 (4.38.0)

### Platform

Microsoft Windows NT 10.0.19045.0 x64
(Windows 10)

### Steps to reproduce

Chat  with several iterations, sometimes with long Copilot answers, at some point scroll appears to be scrolled to the bottom, but there are more messages available. No way to see them.

### Logs (optional)

_No response_

### Expected behavior

I expect to see the entire chat content, especially newer messages, which currently are unavailable.

### Actual behavior

Chat  with several iterations, sometimes with long Copilot answers, at some point scroll appears to be scrolled to the bottom, but there are more messages available. No way to see them.

### Additional information

_No response_

xinyi-gong:
## Summary

Fixes #63. In long conversations the chat could no longer be scrolled down
to the latest messages: the scrollbar looked scrolled to the bottom (or kept
moving) but the content stopped, leaving the newest turns unreachable.

## Root cause

`ChatContentViewer` extended `ScrolledComposite`, which scrolls by moving a
single content child whose height equals the entire conversation. On Windows
a native child window position is a signed 16-bit value, so once the required
offset exceeded that range the OS clamped it — the content could not move
further and the bottom became unreachable. This matches every reported
symptom: only long sessions, Windows only, and "maximize the view" or
"Ctrl+-" (which shrink the total pixel height) working around it.

## Fix

Rewrite `ChatContentViewer` as a self-managed, *windowed* vertical scroller
(`extends Composite` with `SWT.V_SCROLL`):

- `cmpContent` is pinned to the viewport rectangle and never grown or moved.
- A 32-bit logical model (`scrollOffset` / `totalHeight` / per-turn tops)
  drives scrolling; each turn is positioned at `y = top - scrollOffset`.
- Turns intersecting the viewport are laid out; the rest are parked with
  `setVisible(false)`, so no native child ever receives an out-of-range
  coordinate.

Behavior preserved from the previous implementation:

- **Auto-scroll:** scrolling away from the bottom pauses follow-the-tail
  during streaming; scrolling back to the bottom resumes it.
- **Latest user turn pinned to the top:** reserved via virtual bottom
  padding, which also keeps "scroll to bottom" aligned with the real maximum
  so auto-scroll stays correct for both short and long rounds.
- **Streaming stays cheap:** incremental passes re-measure only the trailing
  (mutating) turns; sealed turns keep cached heights.

Scroll step now derives from the OS-provided wheel line count (`event.count`)
times the font line height, so it follows the user's system scroll settings
and DPI instead of a fixed pixel constant.

### DIFF
--- a/com.microsoft.copilot.eclipse.ui.test/src/com/microsoft/copilot/eclipse/ui/chat/ChatContentViewerTest.java
+++ b/com.microsoft.copilot.eclipse.ui.test/src/com/microsoft/copilot/eclipse/ui/chat/ChatContentViewerTest.java
@@ -9,14 +9,17 @@ import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
 import java.lang.reflect.Field;
+import java.lang.reflect.Method;
 import java.util.List;
 import java.util.Map;
 
 import org.eclipse.lsp4j.WorkDoneProgressKind;
 import org.eclipse.swt.SWT;
 import org.eclipse.swt.widgets.Display;
+import org.eclipse.swt.widgets.ScrollBar;
 import org.eclipse.swt.widgets.Shell;
 import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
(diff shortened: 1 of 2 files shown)


jdneo (approved):
LGTM

PR #356, opened 2026-07-17
Add auto model policy support

### LINKED ISSUE #305: Auto Model GA

jdneo:
fix https://github.com/microsoft/copilot-for-eclipse/issues/305

### DIFF
--- a/com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/services/ModelService.java
+++ b/com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/services/ModelService.java
@@ -50,8 +50,6 @@ import com.microsoft.copilot.eclipse.ui.utils.ModelUtils;
  */
 public class ModelService extends ChatBaseService {
 
-  private static final String AUTO_MODEL_ID = "auto";
-
   // models for the model picker
   private IObservableValue<Map<String, CopilotModel>> modelObservable;
   private IObservableValue<CopilotModel> activeModelObservable;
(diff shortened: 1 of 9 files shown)


com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/services/ModelService.java
```diff
@@ -274,31 +281,50 @@ private void updateModelsForChatMode(ChatMode chatMode) {
...
     CopilotModel currentActive = getActiveModel();
     boolean isCurrentModelAvailable = currentActive != null
         && modelsForCurrentMode.containsKey(currentActive.getModelKey());
     if (currentActive == null || !isCurrentModelAvailable) {
       // Try to restore user's preferred model if it's available in current mode
-      String restoredModelId = restoreActiveModel();
-      if (restoredModelId != null && modelsForCurrentMode.containsKey(restoredModelId)) {
-        ensureRealm(() -> activeModelObservable.setValue(modelsForCurrentMode.get(restoredModelId)));
+      String restoredModelKey = restoreActiveModel();
+      if (restoredModelKey != null && modelsForCurrentMode.containsKey(restoredModelKey)) {
```
ethanyhou: From my understanding, the fallback model persistence is skipped by design, right?

As the persisted Auto preference is only checked when the current model is unavailable, after Auto is disabled, the temporary fallback is selected; when Auto is re-enabled, that fallback is still available, so this block is skipped and Auto is never restored.

jdneo: fixed in https://github.com/microsoft/copilot-for-eclipse/pull/356/changes/4aa56fcbc6bba3390ad68ebb725344cf0f8ada9a

ethanyhou (approved):
LGTM

PR #365, opened 2026-07-22
Fix #364 - Use image registries

### LINKED ISSUE #364: [Bug] Images are loaded ad hoc instead of via an image registry
**Summary**
Images across the `ui` and `ui.jobs` bundles are loaded at each call site (mostly through
`UiUtils.buildImageFromPngPath(...)` / `buildImageDescriptorFromPngPath(...)`) instead of through a central
`ImageRegistry`. Each caller passes a hard-coded resource path, gets a freshly created `Image`, caches
it in a field, and is responsible for disposing it. This is repetitive, leak-prone, and hard to maintain.

**Problems**
- **Manual disposal is error-prone and duplicated** — every image owner repeats image instantiation, caching, disposal checks, and registers dispose listeners. Every occurrence is a chance to leak, to access a disposed image, or to double-dispose.
- **Actual resource leaks** — some images are created as static fields and never disposed.
- **Fragile hand-rolled static caches** — several widgets keep a `static Image` field plus a manual
  `disposeStaticIcons()` cleanup that must stay in sync with every read.
- **Duplicate image instances** — `buildImageFromPngPath` decodes and allocates a new `Image` on every call,
  so the same icon is loaded many times instead of being shared/cached.
- **Theme handling duplicated and easy to get wrong** — dark/light selection is re-implemented inline at
  ~17 sites, each branching on `UiUtils.isDarkTheme()` and picking one of two hard-coded paths; inconsistent
  file names (`*_black`/`*_white`, `*_dark`) make it easy to swap the arguments and show the wrong variant.
- **Raw path string literals instead of typed keys** — icons are referenced as strings like
  `"/icons/chat/tools_disabled.png"` (~130+ literals), so typos or moved/renamed resources fail only at
  runtime and there is no single inventory of shipped icons.

**Expected behavior**
Images should be loaded, cached, and disposed using a plug-in's image registry.

travkin79:
This PR fixes issue #364.

- Adds `CopilotImages`, a facade for the ui plug-in's images
- Adds `CopilotJobsImages`, a facade for the jobs plug-in's images
- Adds unit tests to ensure the resources can be loaded
- Improves plug-in instantiation


com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/CopilotImages.java
```diff
@@ -0,0 +1,365 @@
...
+   * Returns the plugin's image registry, initializing it on first access.
+   * Called by {@link CopilotUi#initializeImageRegistry(ImageRegistry)} to pre-register descriptors.
+   */
+  static ImageRegistry getImageRegistry() {
+    ImageRegistry registry = imageRegistry;
+    if (registry == null) {
+      registry = imageRegistry = CopilotUi.getPlugin().getImageRegistry();
+    }
+    return registry;
+  }
```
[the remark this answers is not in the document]
travkin79: fixed

com.microsoft.copilot.eclipse.ui.jobs/src/com/microsoft/copilot/eclipse/ui/jobs/CopilotJobsImages.java
```diff
@@ -0,0 +1,106 @@
...
+   * Returns the plugin's image registry, initializing it on first access.
+   * Called by {@link CopilotJobs#initializeImageRegistry(ImageRegistry)} to pre-register descriptors.
+   */
+  static ImageRegistry getImageRegistry() {
+    ImageRegistry registry = imageRegistry;
+    if (registry == null) {
+      registry = imageRegistry = CopilotJobs.getPlugin().getImageRegistry();
+    }
+    return registry;
+  }
```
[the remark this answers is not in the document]
travkin79: fixed

com.microsoft.copilot.eclipse.ui.test/src/com/microsoft/copilot/eclipse/ui/CopilotUiTests.java
```diff
@@ -10,15 +10,18 @@
...
 import com.microsoft.copilot.eclipse.core.CopilotCore;
 
 class CopilotUiTests {
 
   @Test
   void testCopilotCoreWakeUp() throws Exception {
+    BundleContext context = FrameworkUtil.getBundle(CopilotUi.class).getBundleContext();
     CopilotUi ui = new CopilotUi();
-    ui.start(null);
+    ui.start(context);
```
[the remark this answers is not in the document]
travkin79: fixed

com.microsoft.copilot.eclipse.ui.jobs/META-INF/MANIFEST.MF
```diff
@@ -9,6 +9,7 @@ Bundle-RequiredExecutionEnvironment: JavaSE-17
 Bundle-Localization: plugin
 Automatic-Module-Name: com.microsoft.copilot.eclipse.ui.jobs
 Bundle-ActivationPolicy: lazy
+Export-Package: com.microsoft.copilot.eclipse.ui.jobs
```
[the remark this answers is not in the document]
travkin79: fixed

jdneo:
Thank you, nice refactoring to get rid of tons of dispose listeners.

Give me one more day to do some test on it, if no issues found, I will merge it :)

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/utils/McpUtils.java
```diff
@@ -157,11 +158,11 @@ public static CompletableFuture<Image> loadServerIcon(String iconUrl, int width,
...
   /**
-   * Builds the default MCP server icon image.
+   * Loads the shared default MCP server icon.
    *
-   * @return the default MCP icon image, or {@code null} if it cannot be loaded
+   * @return the default MCP icon image; callers must not dispose it
    */
   public static Image loadDefaultServerIcon() {
-    return UiUtils.buildImageFromPngPath("/icons/mcp/mcp_default_icon.png");
+    return CopilotImages.getImage(CopilotImages.IMG_MCP_DEFAULT_ICON);
```
jdneo, line 169: `loadServerIcon()` now has mixed ownership semantics: this fallback is registry-owned, while both callers (`McpServerItem` and `McpServerDetailDialog`) always dispose the returned image. When the URL is blank or loading fails, disposing the label therefore disposes the shared registry image. Could we make the ownership contract uniform—for example by returning a caller-owned copy, or by handling the registry fallback separately so only downloaded images are disposed?

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/swt/SpinnerAnimator.java
```diff
@@ -88,13 +84,10 @@ public void stop() {
...
     }
-    if (currentFrameImage != null && !currentFrameImage.isDisposed()) {
-      currentFrameImage.dispose();
-    }
     currentFrameImage = null;
   }
 
   private static Image buildFrame(int frame) {
-    return UiUtils.buildImageFromPngPath(String.format("/icons/spinner/%d.png", frame));
+    return CopilotImages.getSpinnerFrame(frame);
```
jdneo, line 89: These frames are now borrowed from the image registry, but the class and `stop()` Javadocs still say that the animator owns, disposes, and frees them. The implementation correctly no longer disposes them; please update the documentation/comments to state that registry frames are borrowed and must never be disposed.

com.microsoft.copilot.eclipse.ui.jobs.test/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,15 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: com.microsoft.copilot.eclipse.ui.jobs.test
+Bundle-SymbolicName: com.microsoft.copilot.eclipse.ui.jobs.test;singleton:=true
+Bundle-Version: 0.20.0.qualifier
+Bundle-Vendor: GitHub Copilot
+Bundle-RequiredExecutionEnvironment: JavaSE-17
+Fragment-Host: com.microsoft.copilot.eclipse.ui.jobs
+Automatic-Module-Name: com.microsoft.copilot.eclipse.ui.jobs.test
+Import-Package: org.objenesis;version="[3.4.0,4.0.0)"
```
jdneo: This test only uses JUnit Assertions and parameterized-test APIs. `org.objenesis`, Mockito Core, and Mockito Jupiter are unused, and requiring `com.microsoft.copilot.eclipse.ui.jobs` is redundant because this bundle already declares it as `Fragment-Host`. Please remove the unnecessary dependencies to keep the test bundle minimal.

com.microsoft.copilot.eclipse.ui.jobs/src/com/microsoft/copilot/eclipse/ui/jobs/CopilotJobs.java
```diff
@@ -17,11 +19,27 @@
  */
 public class CopilotJobs extends AbstractUIPlugin {
 
+  private static volatile CopilotJobs COPILOT_JOBS_PLUGIN;
+
+  /** Returns the shared plugin instance. */
+  public static CopilotJobs getPlugin() {
```
jdneo, line 29: The repository's public-API guideline requires applicable `@param`/`@return` tags. Please add `@return` here and audit the other newly public APIs in both image facades and the `WrappableIconLink.create(...)` overloads, which similarly omit return and/or parameter tags; alternatively, reduce visibility where an API is not intended for consumers.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/CopilotImages.java
```diff
@@ -0,0 +1,358 @@
...
+  public static final String IMG_SPINNER_1 = ICONS_SPINNER + "1.png";
+  public static final String IMG_SPINNER_2 = ICONS_SPINNER + "2.png";
+  public static final String IMG_SPINNER_3 = ICONS_SPINNER + "3.png";
+  public static final String IMG_SPINNER_4 = ICONS_SPINNER + "4.png";
+  public static final String IMG_SPINNER_5 = ICONS_SPINNER + "5.png";
+  public static final String IMG_SPINNER_6 = ICONS_SPINNER + "6.png";
+  public static final String IMG_SPINNER_7 = ICONS_SPINNER + "7.png";
+  public static final String IMG_SPINNER_8 = ICONS_SPINNER + "8.png";
+
+  public static final String[] SPINNER_FRAMES = {
```
jdneo: This exposes mutable registry internals: any consumer can replace an array entry and change which key `getSpinnerFrame()` resolves globally. The eight individual spinner keys also appear to be internal. Could these be private and immutable, with only a frame-count accessor/constant exposed for `SpinnerAnimator`?

jdneo (commented):
Standards/spec review findings. The duplicate ownership finding from both axes is consolidated into one inline comment.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/CopilotImages.java
```diff
@@ -0,0 +1,358 @@
...
+  }
+
+  /**
+   * Returns the theme-correct descriptor for icons that come in light/dark variants.
+   *
+   * @param lightKey key to use in light theme
+   * @param darkKey  key to use in dark theme
+   * @see #getThemedImage(String, String)
+   */
+  public static ImageDescriptor getThemedImageDescriptor(String lightKey, String darkKey) {
```
jdneo: Non-blocking: there are currently no callers for this method, either `getSharedImageDescriptor(...)` implementation, or `CopilotJobsImages.getImageDescriptor(...)`. Could we omit these public hooks until a concrete caller needs them, keeping the new facade API narrower?

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/nes/RulerColumn.java
```diff
@@ -263,7 +264,7 @@ private Rectangle getSuggestionIconBounds(int expectedModelLine, int expectedWid
 
   private void ensureIcon() {
     if (icon == null || icon.isDisposed()) {
-      ImageDescriptor desc = UiUtils.buildImageDescriptorFromPngPath("/icons/chat/gutter-arrow.png");
+      ImageDescriptor desc = CopilotImages.getImageDescriptor(CopilotImages.IMG_CHAT_GUTTER_ARROW);
```
jdneo: Although the descriptor now comes from the registry, the code immediately calls `createImage(true)` below and later manually disposes that new image. Each ruler column therefore still owns a duplicate image, which misses issue #364's caching/ownership goal. Please use `CopilotImages.getImage(IMG_CHAT_GUTTER_ARROW)` directly and remove the corresponding disposal.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/SourceViewerComposite.java
```diff
@@ -184,7 +184,7 @@ private Composite createActionsComposite() {
     ImageDescriptor insertDesc = AbstractUIPlugin.imageDescriptorFromPlugin(UiConstants.WORKBENCH_TEXTEDITOR,
         UiConstants.INSERT_ICON);
     this.insertIcon = Optional.ofNullable(insertDesc).map(desc -> desc.createImage()).orElse(null);
-    Image pasteIcon = PlatformUI.getWorkbench().getSharedImages().getImage(ISharedImages.IMG_TOOL_PASTE);
+    Image pasteIcon = CopilotImages.getSharedImage(ISharedImages.IMG_TOOL_PASTE);
```
jdneo: The adjacent insert icon remains an ad-hoc static-image path: every `SourceViewerComposite` resolves the external descriptor, calls `createImage()`, and disposes the result, while this PR migrates only copy/paste here. For complete issue #364 coverage, could the insert descriptor be registered in an image registry or replaced by a shared image? If the external plug-in exposes no registry-backed option, documenting that exception would clarify why manual ownership remains.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/utils/McpUtils.java
travkin79: You're absolutely right. I'll fix that. I would not recommend to return caller-owned copies, since we would loose caching and other image registry benefits in this case. But I'll return `null` if an image cannot be loaded, instead of returning an image-registry-owned image. Both caller would fall back the the default image themselves and will ensure not to dispose the image-registry-owned image.

com.microsoft.copilot.eclipse.ui.jobs.test/META-INF/MANIFEST.MF
travkin79: I fully agree, I should have had cleaned up the `MANIFEST.MF` more carefully after rebasing on main after PR #361 was merged.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/CopilotImages.java
travkin79: I agree to remove `getSharedImageDescriptor(...)` and `getThemedImageDescriptor(...)` from the image registries. But I suggest to keep `CopilotJobsImages.getImageDescriptor(...)` since it should be preferably used instead of `getImage(...)` to save resources and load the images lazily and IMO it's easier to start using this method if it already exists. I think, all image registries should offer at least the two methods `getImageDescriptor(...)` and `getImage(...)`. Do you agree?

com.microsoft.copilot.eclipse.ui.jobs/src/com/microsoft/copilot/eclipse/ui/jobs/CopilotJobs.java
travkin79: I agree. Would it make sense to let checkstyle complain about missing `@param` and `@return` entries in public API javadoc comments?

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/CopilotImages.java
travkin79: You're right. I'll fix that.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/nes/RulerColumn.java
travkin79: I agree. I somehow missed this case. I'll fix that.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/SourceViewerComposite.java
travkin79: I agree. I'll add the insert icon to the new image registry so that we no longer need to load and must not dispose that icon anymore in the caller, similar to other images taken from the registry.

There is no such shared image that we could easily re-use via `CopilotImages.getSharedImage(...)`.

com.microsoft.copilot.eclipse.ui.jobs/src/com/microsoft/copilot/eclipse/ui/jobs/CopilotJobs.java
jdneo: I'm fine with that, no need to do it in this pr.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/CopilotImages.java
jdneo: Sure, agree on that.

jdneo (commented):
Two more comment:

1. Please check McpRuntimeLogger.java as well, it can use the registry as well.
2. We can remove `UiUtils.resizeIcon(String path, int width, int height)` since no caller now. (not highly related with this PR, but would be great to do the cleanup)

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/tools/RunInTerminalToolAdapter.java
```diff
@@ -199,7 +199,7 @@ public CompletableFuture<LanguageModelToolResult[]> invoke(Map<String, Object> i
       isBackground = Boolean.parseBoolean((String) isBackgroundObj);
     }
 
-    impl.setTerminalIconDescriptor(UiUtils.buildImageDescriptorFromPngPath("/icons/github_copilot.png"));
+    impl.setTerminalIconDescriptor(CopilotImages.getImageDescriptor(CopilotImages.IMG_GITHUB_COPILOT));
```
jdneo: Non-blocking: this centralizes the descriptor lookup, but the terminal migration still stops one step short of issue #364's caching/ownership goal. `IRunInTerminalTool` accepts an `ImageDescriptor`, and both terminal implementations call `createImage()`, cache the resulting Copilot icon, and manually dispose it when idle. That leaves two extra instances of this static image plus hand-managed lifecycle code. Could the SPI instead accept a registry-owned `Image` (or a `Supplier<Image>` resolved on the UI thread), with the implementations borrowing it and never disposing it? Please also account for OSGi lifecycle ordering so terminal tabs cannot retain the image after the UI registry is disposed. The current implementation is not a leak, so I see this as a completeness issue rather than a blocker.

jdneo:
Hi @travkin79 there is a conflict. Please resolve it and I can merge it asap since this is a huge change.

Thank you again for the contribution.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/tools/RunInTerminalToolAdapter.java
travkin79: I fully agree. I didn't see that when migrating the code. I'll adapt the interface to using `Supplier<Image>` and will remove image creation, caching and disposal in the interface implementations.

travkin79:
I rebased, but there is one test failing. I try to fix it quickly.

Update: Fixed now. I think, we're ready to merge. If you want me to squash commits, please let me know.

travkin79:
It would be great if we could merge this PR soon and then go on an finish PRs #368 and #330. The latter is really huge and merge conflicts are likely.

jdneo (approved):

jdneo:
Thank you for your contribution.

PR #363, opened 2026-07-21
feat: make skills generally available

jdneo:
Remove the editor preview feature flag dependency from Skills while retaining the Enable Skills preference.

fix #302

PR #373, opened 2026-07-24
Remove semantic workspace search integration

jdneo:
Remove Eclipse watched-file indexing and project-agent completion while preserving workspace-scoped customizations.

#### Reason for this decision

1. Based on the benchmark, semantic search does not provide significant accurate improvement or lower token consunption
2. There are so many workaround logic to support it in our code base, making code hard to understand
3. It will impact on the startup perf due to the local index.


ethanyhou (approved):
Good clean up. Verified and LGTM

xinyi-gong (approved):
Verified. LGTM

iloveeclipse:
There were no nightly builds for quite some time. Probably good to have one for recent changes like this one or https://github.com/microsoft/copilot-for-eclipse/pull/363?

jdneo:
@iloveeclipse yes, I just kicked off one.

PR #378, opened 2026-07-29
fix: preserve chat cancel state during active turns

### LINKED ISSUE #99: Missing stop button
### Feedback source

None

### GitHub Copilot plugin version

0.15.0.202601270459

### Eclipse IDE and version

Spring Tools for Eclipse  Version: 5.0.1.RELEASE Build Id: 202512192126

### Platform

Darwin Kernel Version 25.3.0: Wed Jan 28 20:53:31 PST 2026; root:xnu-12377.81.4~5/RELEASE_ARM64_T8122 arm64

### Steps to reproduce

Don't know exactly - the stop button just sometimes disappears, so there isn't any way to stop CoPilot.
I think it is usually after it runs an external tool.

### Logs (optional)

_No response_

### Expected behavior

Stop button being there, enabled and working.

### Actual behavior

CoPilot just keeps editing code, and AFAICT I hav no way of stopping it.
I can skip, when it requests to run an external tool, but then it just continues afterward.

### Additional information

_No response_

jdneo:
fix #99

### DIFF
--- a/com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ActionBar.java
+++ b/com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ActionBar.java
@@ -105,7 +105,7 @@ public class ActionBar extends Composite implements NewConversationListener {
   private ContentAssistant ca;
   private Image sendImage;
   private Image sendDisabledImage;
-  private boolean isSendButton = true;
+  private volatile ActionButtonState actionButtonState = ActionButtonState.SEND_DISABLED;
   private LinkedHashSet<MessageListener> messageListeners = new LinkedHashSet<>();
   private Button mcpToolButton;
   private Image mcpToolImage;
@@ -128,10 +128,6 @@ public class ActionBar extends Composite implements NewConversationListener {
   EventHandler featureFlagsChangedEventHandler;
   EventHandler updateMcpToolButtonAndPlaceHolderHandler;
 
-  private static enum SendOrCancelButtonStates {
-    SEND_ENABLED, SEND_DISABLED, CANCEL_ENABLED;
-  }
-
   /**
    * Creates a new InputArea.
    */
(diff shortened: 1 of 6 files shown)


com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatView.java
```diff
@@ -982,9 +982,12 @@ public void onChatProgress(ChatProgressValue value) {
           return;
         }
 
+        boolean isTopLevelTurn = StringUtils.isBlank(value.getParentTurnId());
         if (this.chatContentViewer != null) {
           this.chatContentViewer.processTurnEvent(value);
-          this.actionBar.resetSendButton();
+        }
+        if (isTopLevelTurn) {
+          this.actionBar.markTurnFinished();
```
ethanyhou, line 990: Should we verify that this turn end belongs to the currently active top-level turn before marking it finished? A stale end event from a cancelled turn can arrive after a new turn starts and reset its cancel button.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ActionBar.java
```diff
@@ -731,20 +714,34 @@ private boolean isJavaDebuggerToolEnabledForCurrentMode() {
...
     }
-    this.chatServiceManager.getFileToolService().setWorkingSetBarButtonStatus(true);
+  }
+
+  /**
+   * Returns whether a top-level chat turn is running.
+   *
+   * @return {@code true} while a top-level turn is running
+   */
+  public boolean isTurnRunning() {
```
ethanyhou, line 725: private

jdneo: Used by other type, cannot be private.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/ChatView.java
jdneo: Tracked in https://github.com/microsoft/copilot-for-eclipse/issues/417, will follow in another pr, thanks.

PR #380, opened 2026-07-30
feat: enable MCP sampling support

im-ashu:
Advertise MCP sampling, handle sampling configuration requests, and add a CLI-style inference approval dialog with prompt review and persistent server approval.


im-ashu:

jdneo:
Hi @im-ashu,

Thank you for your contribution. Could you please solve the conflict and mark the review comments as resolved once you pushed the fix commit?

com.microsoft.copilot.eclipse.core/src/com/microsoft/copilot/eclipse/core/lsp/protocol/McpSamplingConfig.java
```diff
@@ -0,0 +1,12 @@
...
+// Licensed under the MIT license.
+
+package com.microsoft.copilot.eclipse.core.lsp.protocol;
+
+import java.util.List;
+
+/**
+ * User preferences for MCP sampling requests from a server.
+ */
+public record McpSamplingConfig(boolean alwaysAllow, boolean alwaysDeny, List<String> allowedModels) {
```
jdneo: I checked the payload from CLS, it might also including

```
val allowedDuringChat: Boolean? = true,
val allowedOutsideChat: Boolean? = true,
```

Are they not being considered in current pr scope?

im-ashu: Good catch, thanks @jdneo! I've pushed a fix that extends `McpSamplingConfig` to include both `allowedDuringChat` and `allowedOutsideChat`.

Since the Eclipse plugin doesn't yet have separate UI/logic to distinguish sampling requests made during an interactive chat session vs. outside of it (e.g. background/agent jobs), both new fields currently mirror `alwaysAllow` -- i.e. if the user has approved "Always Allow" for a server, sampling is allowed in both contexts; otherwise neither is auto-approved and the user is prompted as before.

I added a 5-arg constructor so these can be set independently once/if we add that distinction to the confirmation flow, without touching any of the existing call sites. Let me know if you'd like me to go further and add that context-aware handling in this PR, or if mirroring `alwaysAllow` for now is acceptable to keep the scope focused.

PR #387, opened 2026-08-03
feat: Added Ollama as a BYOK model provider

### LINKED ISSUE #93: Model Management OLLAMA missing
Currently it is not possible in Eclipse Github Copilot in Settings -> Model Management to add OLLAMA. 
In VSCode it is possible.

Can u please add this feature?

ethanyhou:
fix https://github.com/microsoft/copilot-for-eclipse/issues/93

**Test:**
1. Download Ollama latest client app: https://ollama.com/download
2. Pull some models you like
3. Configure the Ollama URL via `Manage Models` preferences page (http://localhost:11434 by default):
   image
5. The configured models should appear in the model selector:
   image
6. Chat with Copilot using Ollama models:
   image


com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/preferences/AddOllamaUrlDialog.java
```diff
@@ -0,0 +1,112 @@
...
+import org.eclipse.swt.widgets.Label;
+import org.eclipse.swt.widgets.Shell;
+import org.eclipse.swt.widgets.Text;
+
+/**
+ * Dialog for configuring the URL shared by Ollama models.
+ */
+public class AddOllamaUrlDialog extends TrayDialog {
+
+  public static final String DEFAULT_ENDPOINT = "http://localhost:11434";
```
jdneo: private?

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/services/ByokService.java
```diff
@@ -137,6 +150,18 @@ public CompletableFuture<Void> loadApiKeys() {
...
+   * Load provider-level endpoint URLs from persistent storage.
+   */
+  public CompletableFuture<Void> loadProviderUrls() {
+    return lsConnection.listByokProviderConfigs(new ByokProviderConfig(null, null)).thenAccept(response -> {
+      Map<String, String> providerUrls = response == null || response.providers() == null ? Map.of()
+          : response.providers().stream().filter(config -> config.url() != null)
+              .collect(Collectors.toMap(ByokProviderConfig::providerName, ByokProviderConfig::url));
+      ensureRealm(() -> providerUrlsObservable.setValue(providerUrls));
+    });
+  }
```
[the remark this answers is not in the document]
ethanyhou: Addressed.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/services/ByokService.java
```diff
@@ -277,12 +368,11 @@ public CompletableFuture<Void> reloadProvider(String providerName) {
   }
 
   /**
-   * Reload all providers sequentially to avoid file write conflicts. Only providers with API keys (excluding Azure)
-   * will fetch remote models.
+    * Reload all providers sequentially to avoid file write conflicts. Providers with API keys and Ollama with a
+    * configured URL will fetch remote models; Azure only uses its locally stored model configurations.
```
jdneo: nit: The comment indentation looks strange. Realign `*`?

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/preferences/ByokPreferencePage.java
```diff
@@ -870,11 +905,31 @@ private boolean showDeleteApiKeyConfirmationDialog(String providerName) {
...
+  }
+
   private void onProviderExpanded(String providerName) {
     // If provider is first expanded, need to fetch models for this provider from remote site
     if (!remotelyLoadedProviders.contains(providerName)) {
-      if (byProviderApiKeys == null || !byProviderApiKeys.containsKey(providerName)) {
-        // No API key for provider, skip loading
+      boolean canFetch = ByokModelProvider.isOllama(providerName) ? byProviderUrls.containsKey(providerName)
+          : byProviderApiKeys.containsKey(providerName);
+      if (!canFetch) {
```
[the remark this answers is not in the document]
ethanyhou: Addressed.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/services/ByokService.java
```diff
@@ -157,7 +186,57 @@ public CompletableFuture<Void> loadLocalModels() {
...
+    });
+  }
+
+  /**
+   * Delete the Ollama URL configuration and all of its stored models.
+   */
+  public CompletableFuture<Void> deleteOllamaConfig() {
+    String providerName = ByokModelProvider.OLLAMA.getDisplayName();
+    return lsConnection.deleteByokProviderConfig(new ByokProviderConfig(providerName, null)).thenCompose(response -> {
+      if (!response.isSuccess()) {
```
jdneo: General question:

At line 206, null check is applied to the response, and here null check is missing. So is the null check necessary?

If no, remove the redundant null check.

If yes, we need to file an issue to track it and fix them separately.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/preferences/AddOllamaUrlDialog.java
```diff
@@ -0,0 +1,112 @@
...
+  private boolean isValidEndpoint(String value) {
+    try {
+      URI uri = new URI(value);
+      return ("http".equalsIgnoreCase(uri.getScheme()) || "https".equalsIgnoreCase(uri.getScheme()))
+          && StringUtils.isNotBlank(uri.getHost());
+    } catch (URISyntaxException e) {
+      return false;
+    }
+  }
+}
```
jdneo:
```suggestion
}

```

jdneo:
### Provider-specific branching is starting to spread

Not a blocker for this PR — the feature works as designed — but I think this is worth capturing before the next provider lands.

`ByokModelProvider.isAzure()` carries this Javadoc:

> *"This avoids scattering direct enum displayName comparisons across UI code."*

That was the right instinct, but the helper only replaced scattered string comparisons with scattered `if` branches. After this PR we're at 13 provider-specific branch sites:

**`ByokService`** — `:334` (`isAzure`, skip remote fetch), `:339` (`isOllama`, check URL instead of key), `:420` (`isOllama`, auto-register discovered models), `:459` (`isAzure`, filter out of bulk fetch)

**`ByokPreferencePage`** — `:519`, `:520`, `:523`, `:525`, `:724`, `:729`, `:848`, `:881`, and the `onProviderExpanded` ternary

All of them are re-deriving answers to the same four orthogonal questions:

1. Does the provider need a **provider-level API key**? (Azure no — per-model credentials; Ollama no; rest yes)
2. Does it need a **provider-level endpoint URL**? (Ollama only)
3. Can it **discover models remotely**? (Azure no)
4. Are discovered models **registered by default**? (Ollama only)

#### The part I'd actually change now

`requiresApiKey()` is the right direction, but it's defined by exclusion:

```java
return !isAzure(providerDisplayName) && !isOllama(providerDisplayName);
```

The next provider that doesn't take a top-level key — and there's demand for one in #93's comments ("OpenAI protocol compatible local LLM server") — silently gets the wrong answer until someone remembers to edit this line. A positive definition would fail loudly instead of quietly.

#### Longer-term shape

Push the capabilities onto the enum so each provider declares itself once:

```java
public enum ByokModelProvider {
  AZURE("Azure", Credential.PER_MODEL, /* remoteDiscovery */ false, /* autoRegister */ false),
  OPENAI("OpenAI", Credential.API_KEY, true, false),
  …
  OLLAMA("Ollama", Credential.ENDPOINT_URL, true, true);

  public boolean requiresApiKey() { … }
  public boolean requiresEndpointUrl() { … }
  public boolean supportsRemoteDiscovery() { … }
  public boolean autoRegistersDiscoveredModels() { … }
}
```

Adding a provider then means one enum line instead of auditing 13 call sites.

I recognise this isn't free: the helpers take `String` because the UI only has the tree's display name, so instance methods would need a `fromDisplayName(String)` lookup plus an unknown-name policy. Happy for that to be a follow-up issue rather than scope in here — but the `requiresApiKey` inversion feels cheap enough to do now.

com.microsoft.copilot.eclipse.core/src/com/microsoft/copilot/eclipse/core/lsp/CopilotLanguageServer.java
```diff
@@ -276,6 +278,24 @@ public interface CopilotLanguageServer extends LanguageServer {
...
+   * Delete a built-in BYOK provider configuration.
+   */
+  @JsonRequest("copilot/byok/deleteProviderConfig")
+  CompletableFuture<ByokStatusResponse> deleteByokProviderConfig(ByokProviderConfig providerConfig);
+
+  /**
+   * List built-in BYOK provider configurations.
+   */
+  @JsonRequest("copilot/byok/listProviderConfigs")
+  CompletableFuture<ByokListProviderConfigResponse> listByokProviderConfigs(ByokProviderConfig providerConfig);
```
jdneo: The param here should be a nullable provider name: https://github.com/microsoft/copilot-language-server-internal/blob/9e251bc3d52b698f364652f7e4dc00fd9bb05ac6/agent/src/methods/byok/listProviderConfigs.ts#L12

com.microsoft.copilot.eclipse.core/src/com/microsoft/copilot/eclipse/core/lsp/CopilotLanguageServer.java
```diff
@@ -276,6 +278,24 @@ public interface CopilotLanguageServer extends LanguageServer {
...
+   * Save a built-in BYOK provider configuration.
+   */
+  @JsonRequest("copilot/byok/saveProviderConfig")
+  CompletableFuture<ByokStatusResponse> saveByokProviderConfig(ByokProviderConfig providerConfig);
+
+  /**
+   * Delete a built-in BYOK provider configuration.
+   */
+  @JsonRequest("copilot/byok/deleteProviderConfig")
+  CompletableFuture<ByokStatusResponse> deleteByokProviderConfig(ByokProviderConfig providerConfig);
```
jdneo: the param here should be a provider name: https://github.com/microsoft/copilot-language-server-internal/blob/9e251bc3d52b698f364652f7e4dc00fd9bb05ac6/agent/src/methods/byok/deleteProviderConfig.ts#L12

ethanyhou:
> ### Provider-specific branching is starting to spread
> Not a blocker for this PR — the feature works as designed — but I think this is worth capturing before the next provider lands.
> 
> `ByokModelProvider.isAzure()` carries this Javadoc:
> 
> > _"This avoids scattering direct enum displayName comparisons across UI code."_
> 
> That was the right instinct, but the helper only replaced scattered string comparisons with scattered `if` branches. After this PR we're at 13 provider-specific branch sites:
> 
> **`ByokService`** — `:334` (`isAzure`, skip remote fetch), `:339` (`isOllama`, check URL instead of key), `:420` (`isOllama`, auto-register discovered models), `:459` (`isAzure`, filter out of bulk fetch)
> 
> **`ByokPreferencePage`** — `:519`, `:520`, `:523`, `:525`, `:724`, `:729`, `:848`, `:881`, and the `onProviderExpanded` ternary
> 
> All of them are re-deriving answers to the same four orthogonal questions:
> 
> 1. Does the provider need a **provider-level API key**? (Azure no — per-model credentials; Ollama no; rest yes)
> 2. Does it need a **provider-level endpoint URL**? (Ollama only)
> 3. Can it **discover models remotely**? (Azure no)
> 4. Are discovered models **registered by default**? (Ollama only)
> 
> #### The part I'd actually change now
> `requiresApiKey()` is the right direction, but it's defined by exclusion:
> 
> ```java
> return !isAzure(providerDisplayName) && !isOllama(providerDisplayName);
> ```
> 
> The next provider that doesn't take a top-level key — and there's demand for one in #93's comments ("OpenAI protocol compatible local LLM server") — silently gets the wrong answer until someone remembers to edit this line. A positive definition would fail loudly instead of quietly.
> 
> #### Longer-term shape
> Push the capabilities onto the enum so each provider declares itself once:
> 
> ```java
> public enum ByokModelProvider {
>   AZURE("Azure", Credential.PER_MODEL, /* remoteDiscovery */ false, /* autoRegister */ false),
>   OPENAI("OpenAI", Credential.API_KEY, true, false),
>   …
>   OLLAMA("Ollama", Credential.ENDPOINT_URL, true, true);
> 
>   public boolean requiresApiKey() { … }
>   public boolean requiresEndpointUrl() { … }
>   public boolean supportsRemoteDiscovery() { … }
>   public boolean autoRegistersDiscoveredModels() { … }
> }
> ```
> 
> Adding a provider then means one enum line instead of auditing 13 call sites.
> 
> I recognise this isn't free: the helpers take `String` because the UI only has the tree's display name, so instance methods would need a `fromDisplayName(String)` lookup plus an unknown-name policy. Happy for that to be a follow-up issue rather than scope in here — but the `requiresApiKey` inversion feels cheap enough to do now.

Addressed in https://github.com/microsoft/copilot-for-eclipse/issues/388

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/preferences/AddOllamaUrlDialog.java
ethanyhou: Updated

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/services/ByokService.java
ethanyhou: fixed

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/chat/services/ByokService.java
ethanyhou: removed

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/preferences/AddOllamaUrlDialog.java
ethanyhou: updated

com.microsoft.copilot.eclipse.core/src/com/microsoft/copilot/eclipse/core/lsp/CopilotLanguageServer.java
ethanyhou: Updated

com.microsoft.copilot.eclipse.core/src/com/microsoft/copilot/eclipse/core/lsp/CopilotLanguageServer.java
ethanyhou: updated

jdneo (approved):

PR #392, opened 2026-08-05
fix: prevent test processes from hanging CI

jdneo:
## Summary

M2/3 (bottom) of the CI-hang stack. M1's exception-deadlock fix and M3's macOS shell-reader fix follow as separate layers. Because this intermediate PR intentionally excludes those fixes, CI may still hang until the full stack is applied.

This four-file layer:

- caps CI jobs at 45 minutes in `.github/workflows/ci.yml`
- caps inactive forked test JVMs at 900 seconds in the root `pom.xml`
- skips no-op SWTBot workbench launches unless `-Dprobe.script` is supplied
- makes the language-server test expose, stop, forcibly terminate if necessary, await, and assert exit of its spawned process

## Validation

- `./mvnw checkstyle:check` passes
- `./mvnw test` passes across all 19 modules, including the SWTBot no-op path
- `./mvnw clean verify` reaches compilation but packaging is blocked by the checkout's pre-existing missing `com.microsoft.copilot.eclipse.core/copilot-agent/dist/` input
- the four-file diff exactly matches the corresponding snapshot hunks from `01cf6650bf537bf3eb5c44f442b44ecbdd7353e3`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,6 +13,9 @@ jobs:
   build:
 
     runs-on: ${{ matrix.os }}
+    # Without an explicit limit a hung test JVM keeps the job alive for the 6 hour default and the
+    # logs are never published, which makes such hangs impossible to diagnose.
+    timeout-minutes: 45
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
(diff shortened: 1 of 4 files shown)


jdneo:
ubuntu fail is as expected since it is force terminated when timeout

ethanyhou (approved):

PR #393, opened 2026-08-05
fix: avoid exception reporting deadlocks

jdneo:
M1/3 middle stack layer. Depends on #392 (M2); M3 follows with the macOS shell-reader fix. This intermediate layer may still hang in CI until M3 lands.

This 12-file layer replaces synchronous framework exception reporting with a bounded asynchronous reporter, avoids language-server wrapper re-entry during telemetry, bounds initialization and shutdown lifecycle operations, moves telemetry frame filename construction into `TelemetryExceptionParams`, and adds focused exception reporter, connection lifecycle, and telemetry protocol tests.

Validation:
- `./mvnw checkstyle:check`
- `./mvnw test`
- `./mvnw clean verify -Dtycho.testArgLine=-Dnet.bytebuddy.experimental=true`


com.microsoft.copilot.eclipse.core/src/com/microsoft/copilot/eclipse/core/CopilotCore.java
```diff
@@ -65,40 +85,97 @@ public static CopilotCore getPlugin() {
...
+      try {
+        if (job != null) {
+          job.join();
+        }
+      } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
+      } finally {
+        stopLanguageServer();
+      }
+    }, "Copilot shutdown");
```
ethanyhou, line 118: If the initialization job is blocked while starting the language server, `job.join()` will never complete, so `stopLanguageServer()` will never be reached. The 30-second timeout allows `CopilotCore.stop()` to return, but the initialization job or language-server process may still be running in the background.

ethanyhou (approved):

PR #394, opened 2026-08-05
fix: stop the macOS login shell from pinning the JVM

jdneo:
M3/3 top stack layer.

Depends on #393 and #392. Full stack CI is authoritative.

This one-file change prevents the macOS login-shell environment reader from pinning the JVM by using a named daemon reader thread and discarding the otherwise-undrained shell stderr.

Validated with Checkstyle, tests, and clean verify. On local Java 25, clean verify used `-Dtycho.testArgLine=-Dnet.bytebuddy.experimental=true` for the known Mockito/Byte Buddy compatibility requirement.

### DIFF
--- a/com.microsoft.copilot.eclipse.core/src/com/microsoft/copilot/eclipse/core/lsp/LsStreamConnectionProvider.java
+++ b/com.microsoft.copilot.eclipse.core/src/com/microsoft/copilot/eclipse/core/lsp/LsStreamConnectionProvider.java
@@ -332,6 +332,9 @@ public class LsStreamConnectionProvider extends ProcessStreamConnectionProvider 
     try {
       // Execute login shell and get environment variables
       ProcessBuilder pb = new ProcessBuilder("/bin/zsh", "-i", "-l", "-c", "env");
+      // Nobody drains the shell's stderr, so a chatty shell profile would otherwise fill the pipe
+      // buffer and wedge the shell before it ever prints the environment.
+      pb.redirectError(ProcessBuilder.Redirect.DISCARD);
       process = pb.start();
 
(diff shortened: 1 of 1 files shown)


com.microsoft.copilot.eclipse.core/src/com/microsoft/copilot/eclipse/core/lsp/LsStreamConnectionProvider.java
```diff
@@ -359,8 +362,14 @@ private Map<String, String> getLoginShellEnvironment() {
...
-    // Create a separate thread to read the process output with a timeout, this avoids blocking the original thread
-    ExecutorService executor = Executors.newSingleThreadExecutor();
+    // Reads on a process pipe are not interruptible, so neither cancel(true) nor shutdownNow() can
+    // stop the reader once the shell stops producing output. The thread must therefore be a daemon,
+    // otherwise a wedged shell keeps the whole JVM alive.
+    ExecutorService executor = Executors.newSingleThreadExecutor(runnable -> {
+      Thread thread = new Thread(runnable, "GitHub Copilot login shell environment reader");
+      thread.setDaemon(true);
+      return thread;
+    });
```
ethanyhou, line 372: From Copilot, optional: Making this reader thread a daemon prevents it from pinning JVM shutdown, but it also means a timed-out reader may remain alive for the rest of the Eclipse session if the shell does not exit after process.destroy(), or if a descendant keeps stdout open. Could we make the timeout cleanup deterministic by closing the streams, waiting for the process, and falling back to descendant cleanup plus destroyForcibly()?

ethanyhou (approved):

PR #415, opened 2026-08-12
feat: add Copilot Java quick assist

### LINKED ISSUE #70: Please integrate copilot into eclipse quickfix feature - Enhancement Request
Real world example:

I have some sonar warnings ...

Image

Quickfix (Ctrl + 1) shows ...

Image

Why not adding "Suggest fix from copilot" here too?

- the code-context ist known
- the (external) problem is available (eclipse problem marker(s)

So it should be easy to feed copilot with that data and ask for a focused fix .. only this line 

In my example it would be pretty cool if copilot can suggest to use correct slf4j logging pattern (maybe he can also gain some more input from the referenced variables, the surrounding method/class, or even more (performance?) ... 

A fix could be to do "LOGGER.warn("Ignore action '{}'", action);"

lytreallynb:
## Summary

- Add a JDT quick assist that offers **Suggest fix with GitHub Copilot** for problem markers overlapping the caret or current selection.
- Deduplicate diagnostics in source order and prefill the existing Copilot Chat flow without automatically sending the prompt.
- Hide the proposal while Copilot is signed out or still initializing.

This intentionally limits the initial implementation to `org.eclipse.jdt.ui.quickAssistProcessors`, as agreed in the issue discussion.

## Testing

- `mvn -q verify -Dtest=QuickFixProcessorSupportTests,QuickAssistExtensionTests -DfailIfNoTests=false`
- `mvn -q clean verify`
- Manually verified in Eclipse 2026-06 that the proposal is hidden while signed out, appears for a Java problem marker when signed in, attaches the active Java file, prefills the diagnostic prompt, and does not auto-send.

Fixes #70


lytreallynb:

jdneo:
Thanks, please check the ci failure. Meanwhile some questions:

1. I noticed that when the quick fix is selected, the affected range will be automatically selected, but I did not find such logic in the PR, is this controlled by upstream jdt.ui? I mean, if it's `jdt.ui`'s behavior, we have potential regression in the future if upstream changes its behavior.

2. What if a range has multiple problem marks? What will happen?

lytreallynb:
Thanks, these were good catches.

I checked JDT's `JavaCorrectionAssistant` implementation. It can temporarily move the caret to the nearest quick-fixable annotation with a zero-length selection, and it restores the original selection when the proposal popup closes. So yes, the behavior we observed comes from JDT UI, but it does not provide a stable, non-empty affected range that this integration should rely on.

I pushed 4877ace to make the context explicit. The Copilot proposal now returns its own selection from `getSelection()`. For a caret invocation, that selection covers the matched diagnostic range, or the union of the ranges when multiple matching markers overlap the caret. If the user invokes Quick Assist with an existing non-empty selection, that selection is preserved.

For multiple problem markers at the same range, we show one Copilot proposal. Its prompt contains all unique diagnostic messages, ordered by marker offset and then by message for deterministic ties. Identical messages are deduplicated. I added a regression test for multiple markers sharing the same range, including a duplicate message.

I also checked the CI failure. The Java 17 test runtime was instantiating the processor, which caused it to load a JDT UI interface compiled for Java 21. The extension registration test now verifies the registry metadata without loading the processor class. A complete `clean verify` passes locally with the project's Java 21 target runtime, including 10,206 core tests, 468 UI tests, and 4 UI jobs tests.

com.microsoft.copilot.eclipse.ui/src/com/microsoft/copilot/eclipse/ui/quickfix/JavaCopilotQuickAssistProcessor.java
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  JavaCopilotQuickAssistProcessor(BooleanSupplier isCopilotAvailable) {
+    this.isCopilotAvailable = isCopilotAvailable;
+  }
+
+  @Override
+  public boolean hasAssists(IInvocationContext context) throws CoreException {
+    return createProposal(context) != null;
+  }
```
jdneo, line 51: Could `hasAssists()` use a lightweight availability check instead of creating the full proposal? JDT may call this method frequently when updating the quick-assist light bulb. Currently, `createProposal()` copies the entire compilation unit into a new `Document`, scans all problem markers, sorts and deduplicates them, builds the prompt, and creates a proposal. The same work is then repeated by `getAssists()`.

Consider adding a short-circuiting helper that returns as soon as it finds an overlapping problem marker, while leaving the full proposal construction to `getAssists()`.

lytreallynb: Thanks for pointing this out. I updated `hasAssists()` to use a separate availability check that stops as soon as it finds an overlapping problem marker. `getAssists()` still handles the full diagnostic collection and proposal construction, so that work is no longer repeated just to update the light bulb. I added tests for the new path, and `clean verify` passes locally.

The availability check still creates a `Document` up front because we also support line-only markers. I left that unchanged for now to keep this update focused. If you think it would be worthwhile, I can also make the `Document` creation lazy for the usual character-range marker case.

jdneo (approved):
LGTM. Thank you for your contribution.

PR #448, opened 2026-09-15
Fix workspace folder URI resolution for SCM-backed projects

HHartvig:
## Problem
`WorkspaceUtils.listWorkspaceFolders()` used `project.getLocationURI()` directly to
build LSP `WorkspaceFolder` entries. For projects backed by a source-control provider
(e.g. `semanticfs://`, `sourcecontrol://`), this could return a URI that didn't match
what the rest of the codebase expects, since it bypassed the same file-URI resolution
used elsewhere (`LSPEclipseUtils.toUri()` + platform-resource fallback).
This made add custom agents fail, making it impossible to add custom agents on non git repos. 

## Solution
- Added `FileUtils.getResourceAsUri(IResource)`, a resource-agnostic version of the
  existing URI resolution logic: prefers `LSPEclipseUtils.toUri()` when the resource
  has a local `getLocation()`, falls back to `getLocationURI()` for remote/virtual
  resources, and finally falls back to a `platform:/resource...` URI built from the
  resource's portable full path.
- Updated `WorkspaceUtils.listWorkspaceFolders()` to use `FileUtils.getResourceAsUri()`
  instead of calling `project.getLocationURI()` directly, so workspace folders are
  resolved consistently with the rest of the codebase regardless of the project's
  backing storage/SCM provider.

## Testing
Added unit tests in `FileUtilsTests` covering all resolution branches of
`getResourceAsUri`:
- Resource with a local `getLocation()` → returns the `LSPEclipseUtils`-resolved URI
- Resource where `getLocation()` resolves to a `file://` URI derived from the absolute
  path, even when `getLocationURI()` reports a different SCM scheme (locally-backed
  resource takes priority)
- Resource without `getLocation()` but with `getLocationURI()` → returns that URI directly
- Resource without either → falls back to a `platform:/resource...` URI

Ran `.\mvnw test` — all tests pass.


HHartvig:
@microsoft-github-policy-service agree company="Danske Bank"

